PR #211, opened 2017-10-05
vendoring rbac

puja108:


calvix:
I would wait with merging because atm if a master is recreated updates are applied. This enables RBAC on clusters and broke lot of stuff for customer

calvix:
legacy PR, we now have some workaround and merge will be done after all customers move to RBAC

PR #216, opened 2017-10-13
Create encryption config

jgsqware:


service/cloudconfig/v_0_1_0_master_template.go
```diff
@@ -137,6 +140,12 @@ func (e *v_0_1_0MasterExtension) Files() ([]k8scloudconfig.FileAsset, error) {
 			Owner:        FileOwner,
 			Permissions:  FilePermission,
 		},
+		{
+			AssetContent: e.keys.APIServerEncryptionKey,
+			Path:         "/etc/kubernetes/encryption/k8s-encryption-config.yaml.enc",
+			Owner:        FileOwner,
+			Permissions:  0644,
```
r7vme: please use more strict permissions e.g. 0600

service/cloudconfig/v_0_1_0_master_template.go
```diff
@@ -137,6 +140,12 @@ func (e *v_0_1_0MasterExtension) Files() ([]k8scloudconfig.FileAsset, error) {
 			Owner:        FileOwner,
 			Permissions:  FilePermission,
 		},
+		{
+			AssetContent: e.keys.APIServerEncryptionKey,
+			Path:         "/etc/kubernetes/encryption/k8s-encryption-config.yaml.enc",
```
r7vme: as far as i understand this file contains NOT encrypted base64 encoded key in yaml. So no need for `.enc` extension.

jgsqware: Good catch. Copy/Paste error

service/cloudconfig/v_0_1_0_master_template.go
jgsqware: Ok. I've mapped to the config file permission

```
ls -lrth /etc/kubernetes/config/
total 16K
-rw-r--r--. 1 root root 392 Oct 12 12:05 proxy-kubeconfig.yml
-rw-r--r--. 1 root root 396 Oct 12 12:05 kubelet-kubeconfig.yml
```

r7vme: kubeconfig itself is ok, as it does not contain (in our) any private keys. But i see that even private keys have insecure permissions on host cluster.

```
r7vme@ip-10-0-5-126 ~ $ ls -la /etc/kubernetes/ssl
total 64
drwxr-xr-x. 4 root root 4096 Sep 13 19:58 .
drwxr-xr-x. 6 root root 4096 Sep 13 19:59 ..
-rw-r--r--. 1 root root 1219 Sep 13 19:58 apiserver-ca.pem
-rw-r--r--. 1 root root 1459 Sep 13 19:58 apiserver-crt.pem
-rw-r--r--. 1 root root 1674 Sep 13 19:58 apiserver-key.pem
drwxr-xr-x. 2 root root 4096 Sep 13 19:56 calico
drwxr-xr-x. 2 root root 4096 Sep 13 19:56 etcd
-rw-r--r--. 1 root root 1651 Sep 13 19:58 service-account-key.pem
```

This means that anyone who has access to node will be able to read private key.

btw, i've checked guest clusters and they have correct permissions
```
r7vme@ip-10-0-5-126 ~ $ ls -la /etc/kubernetes/ssl/etcd/
total 64
drwxr-xr-x. 2 root root 4096 Sep 13 19:56 .
drwxr-xr-x. 4 root root 4096 Sep 13 19:58 ..
-rw-r--r--. 1 root root 1219 Sep 13 19:56 client-ca.pem
-rw-r--r--. 1 root root 1333 Sep 13 19:56 client-crt.pem
-rw-r--r--. 1 root root 1678 Sep 13 19:56 client-key.pem
-rw-r--r--. 1 root root 1219 Sep 13 19:56 server-ca.pem
-rw-r--r--. 1 root root 1333 Sep 13 19:56 server-crt.pem
-rw-r--r--. 1 root root 1678 Sep 13 19:56 server-key.pem
```

So far you need to use strict permissions. And we also need to fix them for [host clusters](https://github.com/giantswarm/giantswarm/issues/1680#issuecomment-336806139).

r7vme: After discussion here https://github.com/giantswarm/giantswarm/issues/1680

makes sense to just use default strict `Permissions:  FilePermission,`

calvix: yep for sure lets use 600 for the encrypt key

r7vme:
When this will be done please make sure to enable encryption config for kvm-operator here https://github.com/giantswarm/k8scloudconfig/commit/6fcdf96d070a64106b38f68a45891c7959fe0e20

r7vme:
Any progress so far? This is one of the dependencies (along with RBAC) to use k8scc from master.

cc: @calvix @corest

xh3b4sd:
What happens with this PR? Can we merge or close it? I would like to keep the repo clean and get rid of such orphaned things.

jgsqware:
Not dead and not orphan. It's mine but no had time to work on it so far. I
will try to get rid of it next week.
Ok?

xh3b4sd:
Ack, thanks for the info. <3

r7vme:
@jgsqware thanks, hope to see this :) then we can get rid of separate branch for k8scc for on-prem

jgsqware:
@xh3b4sd @calvix @r7vme I've mergeed back master and tested it on Geckon and encryption works.

WDYT?

service/resource/configmap/desired.go
```diff
@@ -39,8 +40,27 @@ func (r *Resource) newConfigMaps(customObject kvmtpr.CustomObject) ([]*apiv1.Con
 		return nil, microerror.Mask(err)
 	}
 
+	keys, err := r.keyWatcher.SearchKeys(customObject.Spec.Cluster.Cluster.ID)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
```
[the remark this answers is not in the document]
jgsqware: Good point, I like that, I've done it what do you think?

service/resource/configmap/desired.go
```diff
@@ -39,8 +40,27 @@ func (r *Resource) newConfigMaps(customObject kvmtpr.CustomObject) ([]*apiv1.Con
...
+		return nil, microerror.Maskf(notFoundError, "could not get encryption keys from secrets")
+	}
+
+	encryptionConfig, err := r.EncryptionConfig(string(encryptionKey))
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	compactKeys := randomkeytpr.CompactRandomKeyAssets{
+		APIServerEncryptionKey: encryptionConfig,
```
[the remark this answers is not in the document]
jgsqware: true
I'll create a boyscouting issue for that

service/cloudconfig/cloud_config.go
```diff
@@ -52,7 +53,7 @@ func New(config Config) (*CloudConfig, error) {
 
 // NewMasterTemplate generates a new worker cloud config template and returns it
 // as a base64 encoded string.
-func (c *CloudConfig) NewMasterTemplate(customObject kvmtpr.CustomObject, certs certificatetpr.AssetsBundle, node clustertprspec.Node) (string, error) {
+func (c *CloudConfig) NewMasterTemplate(customObject kvmtpr.CustomObject, certs certificatetpr.AssetsBundle, node clustertprspec.Node, keys map[randomkeytpr.Key][]byte) (string, error) {
```
xh3b4sd: Here we see how inconsistent the implementation of the keys is. We have a kind of clean type `certificatetpr.AssetsBundle` for the certs. It would be nice to have something similar for the keys. Lets create an issue for this to refactor at some point.

service/cloudconfig/v_0_1_0_master_template.go
```diff
@@ -38,9 +45,19 @@ func v_0_1_0MasterTemplate(customObject kvmtpr.CustomObject, certs certificatetp
 
 type v_0_1_0MasterExtension struct {
 	certs certificatetpr.AssetsBundle
+	keys  []byte
```
xh3b4sd: This looks like `keys` is now only one key. IMO it should be what we get just like with the certs.

service/cloudconfig/v_0_1_0_master_template.go
```diff
@@ -38,9 +45,19 @@ func v_0_1_0MasterTemplate(customObject kvmtpr.CustomObject, certs certificatetp
...
 
 func (e *v_0_1_0MasterExtension) Files() ([]k8scloudconfig.FileAsset, error) {
+	encryptionConfig, err := EncryptionConfig(string(e.keys))
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	compactKeys := randomkeytpr.CompactRandomKeyAssets{
+		APIServerEncryptionKey: encryptionConfig,
+	}
```
xh3b4sd: Is this now still necessary? The encryption config can just be set below right?

service/cloudconfig/cloud_config.go
jgsqware: Will make it

service/cloudconfig/v_0_1_0_master_template.go
jgsqware: you mean something like `encryptionKey` instead

xh3b4sd: I mean there is a `certs` argument given which has the type `certificatetpr.AssetsBundle`. This is applied to the `v_0_1_0MasterExtension` type as it is. For the key you do it differently and I wonder why. When type A is given we should probably just use type A and not type B.

jgsqware: Ok got you. So it would be preferable to validate the map, but retrieving the key in `.Files()` when creating `EncryptionConfig`

I'll do that

xh3b4sd: Yeah. Cool.

jgsqware: ok done

service/resource/configmap/desired.go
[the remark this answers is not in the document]
xh3b4sd: Yeah this is nicer.

calvix:
IS the key created even for guest cluster created later? if not we have a problem where old clusters without the key will break.

xh3b4sd:
Of course they are not. So this should be made backwards compatible before merging, like any change we release should be.

jgsqware:
Good point, No it will create only for new cluster
I have no clue how to do this?

xh3b4sd:
We do not have a mechanism to "fix" existing clusters that way right now. I have something on my list to achieve this but this will not land tomorrow and will take a bit time. This means changes being introduced have to be backwards compatible. In case the encryption key is not there, the reconciliation should respect this for the transition period.

calvix:
I did some fixes by manually updating the TPRs as that was the fastes solution and only took like 15 mins total

r7vme:
+1 on @xh3b4sd point.

We need respect old clusters, so all new clusters by default should have some anchor that we can rely on in parametrizing encryption.

jgsqware:
@xh3b4sd @kopiczko Updated for versioning. I've added the encryption in `cloudconfigv2`, using the `k8scloudconfig` `V2.0.0`

service/cloudconfigv2/cloud_config.go
```diff
@@ -51,32 +53,26 @@ func New(config Config) (*CloudConfig, error) {
...
-			certs: certs,
-		}
-		params.Node = node
-	}
-
-	var newCloudConfig *k8scloudconfig.CloudConfig
-	{
-		newCloudConfig, err = k8scloudconfig.NewCloudConfig(k8scloudconfig.MasterTemplate, params)
+	var template string
+	switch customObject.Spec.Cluster.Version {
```
xh3b4sd: This does not work. The cluster version field was actually removed. Ross added it back because of some details in AWS. Versioning has to happen via version bundles and resource routing. IMO the cloud config package should stay as it is. We should rather introduce another one which does the magic, which implies that another configmap resource packages has to be introduced then as well. Otherwise we are either not able to version or screw existing clusters like before.

jgsqware: Ok I would need a deeper explanatoin on version bundle and resource routing so

xh3b4sd: Check 

- https://github.com/giantswarm/aws-operator/blob/master/service/resource_router.go
- https://github.com/giantswarm/aws-operator/blob/master/service/version_bundles.go

Note that the resource routing is not yet implemented so this should happen as a first step in a separate PR. 

Another PR should then be to copy the cloudconfig package and the configmap resource package. 

Another PR should then be to add changes to the copied packages accordingly. 

Another PR should then be to add the new resource to the resource router.

jgsqware:
Closing in favor of https://github.com/giantswarm/kvm-operator/pull/320

xh3b4sd:
Is there a reason for the branch still being there? Can we remove it?

PR #229, opened 2017-11-01
Add iscsid/multipathd services for the onprem installation

corest:


vendor/github.com/giantswarm/k8scloudconfig/v_0_1_0/master_template.go
```diff
@@ -1789,6 +1796,10 @@ coreos:
       --v=2"
       ExecStop=-/usr/bin/docker stop -t 10 $NAME
       ExecStopPost=-/usr/bin/docker rm -f $NAME
+  - name: iscsid.service
```
calvix: there seems to be vendored the old version :)

corest:
Works fine with system services as well

PR #231, opened 2017-11-02
vendor-k8scc-etcd-localhost

calvix:
vendor k8scc changes that will let k8s-api use localhost to reach etcd

towards https://github.com/giantswarm/giantswarm/issues/1905

https://github.com/giantswarm/kvm-operator/blob/f95eee072e551f50973f4291e0018a76432c0c1c/vendor/github.com/giantswarm/k8scloudconfig/v_0_1_0/master_template.go#L1941


calvix:
tested on lycan, works  as expected

PR #325, opened 2017-12-20
initial keyv3 and deploymentv3 (v2 copies)

fgimenez:
Towards #259


fgimenez:
Packages just copied, merging.

PR #342, opened 2018-01-12
add-empty-annotations

calvix:
towards https://github.com/giantswarm/giantswarm/issues/2359
As in new k8s we cant update pod manifest on the fly we need to add the annotations when the pod is created and then later we just update the value.


calvix:
Done, tested on gorgoth. endpoints are updated even on new version of k8s.

update is done via Patch instead

PR #351, opened 2018-01-23
Updates vendor with k8s 1.9.2

corest:


kopiczko:
What is going on here? Assigned to @corest. @pipo02mix is pushing stuff. Do we need to move some stuff from there to v4? Can it be closed?

corest:
outdated

PR #357, opened 2018-01-30
kvmconfig/v3: copy missing resources from v2

kopiczko:
This is pure copy. Nothing changed.


kopiczko:
Yeah, thanks! I don't have an issue with opening new ones though.

PR #383, opened 2018-02-07
kvmconfig/v3: remove gauge init

kopiczko:
This causes panics in v4.

I know it's change on frozen version. But I don't see how we can continue
otherwise. I'll propose a solution which gets rid of the problem.


kopiczko:
Yes, but anyway, we should try to avoid that. Especially when we have the tooling preventing that.

PR #385, opened 2018-02-08
kvmconfig: extract metric package

kopiczko:
This prevents future panics when copying packages.


service/kvmconfig/metric/metric.go
```diff
@@ -1,16 +1,15 @@
-package deployment
+package metric
```
kopiczko: I had to extract a new package  to avoid cycles. Still `metric.VersionBundleVersionGauge` reads nice.

service/kvmconfig/metric/metric.go
```diff
@@ -0,0 +1,24 @@
...
```
kopiczko: I had to create a new package to avoid cycles. Still `metric.VersionBundleVersionGauge` reads nice and package makes sense (at least for me :) ).

service/kvmconfig/v4/key/key.go
```diff
@@ -47,6 +47,8 @@ const (
 	// kvm endpoint annotations
 	AnnotationIp      = "endpoint.kvm.giantswarm.io/ip"
 	AnnotationService = "endpoint.kvm.giantswarm.io/service"
+
+	VersionBundleVersionAnnotation = "giantswarm.io/version-bundle-version"
```
kopiczko: Moved to `key` package where I believe it belongs.

service/kvmconfig/v4/resource/deployment/metric.go
```diff
@@ -1,25 +0,0 @@
-package deployment
```
kopiczko: moved to kvmconfig/metric/metric.go

service/kvmconfig/metric/metric.go
tuommaki: I think this is good for now, but if we have similar gauges more in projects, would it make sense to extract these to a library where you could just issue them with one function call?

kopiczko: I didn't think about that. Right now it's a single metric. I'd say we should start to think about it when we have ~10 common metrics. One of the problem is that namespace change, but probably not very challenging.

PR #394, opened 2018-02-19
increase-init-delay-node-controller

calvix:
increase liveness probe init delay

towards https://github.com/giantswarm/kvm-operator/issues/344

PR #399, opened 2018-02-20
update-kvm-node-controller

calvix:
instead of https://github.com/giantswarm/kvm-operator/pull/395 and https://github.com/giantswarm/kvm-operator/pull/394

* updating kvm node controller to the latest version with pod status bugfix - https://github.com/giantswarm/kvm-operator-node-controller/issues/12
* increasing initial health check for node controller

### DIFF
--- a/service/version_bundles.go
+++ b/service/version_bundles.go
@@ -292,5 +292,50 @@ func NewVersionBundles() []versionbundle.Bundle {
 			Version:      "2.0.0",
 			WIP:          false,
 		},
+		{
+			Changelogs: []versionbundle.Changelog{
+				{
+					Component:   "kvm-node-controller",
+					Description: "Updated KVM node controller with pod status bugfix.",
+					Kind:        versionbundle.KindChanged,
+				},
+			},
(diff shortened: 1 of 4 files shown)


service/version_bundles.go
```diff
@@ -292,5 +292,50 @@ func NewVersionBundles() []versionbundle.Bundle {
 			Version:      "2.0.0",
 			WIP:          false,
 		},
+		{
+			Changelogs: []versionbundle.Changelog{
+				{
+					Component:   "kvm-node-controller",
+					Description: "Updated KVM node controller with pod status bugfix.",
+					Kind:        versionbundle.KindChanged,
+				},
```
kopiczko: So, should we also list it in Components?

calvix: thing is, we don't have "versions" for our components, they are just tags
Also, I am not sure how much interesting information is that for the customer.

kopiczko: Yes this is weird situation. Something to think about. Not interesting but we have that in the changelog which is customer facing.

PR #410, opened 2018-02-26
update to latest k8scc

calvix:
update to latest k8scc

### DIFF
--- a/service/kvmconfig/v6/version_bundle.go
+++ b/service/kvmconfig/v6/version_bundle.go
@@ -19,6 +19,16 @@ func VersionBundle() versionbundle.Bundle {
 				Description: "Updated to 3.0.2.",
 				Kind:        versionbundle.KindChanged,
 			},
+			{
+				Component:   "kubelet",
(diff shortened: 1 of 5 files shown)


service/kvmconfig/v6/version_bundle.go
```diff
@@ -19,6 +19,16 @@ func VersionBundle() versionbundle.Bundle {
...
+			{
+				Component:   "kubelet",
+				Description: "Tune kubelet flags for protecting key units (kubelet and container runtime) from workload overloads.",
+				Kind:        versionbundle.KindChanged,
+			},
+			{
+				Component:   "etcd",
+				Description: "Updated to 3.3.1.",
+				Kind:        versionbundle.KindChanged,
+			},
```
kopiczko: I thought for guest clusters, we use etcd that comes with the Containter Linux?

calvix: No, not for a long time.

kopiczko: Oh yeah, I confused it with docker.

PR #411, opened 2018-02-26
redesign-qemu-overhead-calculation

calvix:
redesign-qemu-overhead-calculation

Now we calculate overhead based on memory and also include 512M overhead for IO operations.

For masters, I left the calculation same as before that is +1GB as atm their workload run quite differently and we should be fine there anyway.

new formula calculates the limits in MB as GB is too big and we would waste many memory in case small clusters.


service/kvmconfig/v6/key/key.go
```diff
@@ -124,18 +128,54 @@ func MasterHostPathVolumeDir(clusterID string, vmNumber string) string {
...
+		// memory between 30 - 45G >> overhead 2048M
+		overheadMultiplier := int(baseWorkerOverheadMultiplier + i/baseWorkerOverheadModulator)
+		workerMemoryOverhead := strconv.Itoa(baseWorkerMemoryOverhead*overheadMultiplier) + "M"
+
+		memOverhead, err := resource.ParseQuantity(workerMemoryOverhead)
+		if err != nil {
+			return resource.Quantity{}, microerror.Maskf(err, "creating Memory quantity from memory overhead")
+		}
+		q.Add(memOverhead)
 	}
```
kopiczko: It would be more type safe to split it to two separate functions like `MemoryQuantityMaster` and `MemoryQuantityWorker`. When we keep that `if` the code can fail silently because of a typo.

service/kvmconfig/v6/key/key.go
```diff
@@ -124,18 +128,54 @@ func MasterHostPathVolumeDir(clusterID string, vmNumber string) string {
...
+// It adds the memory from the node definition parameter to the additional memory calculated on the node role
+func MemoryQuantity(n v1alpha1.KVMConfigSpecKVMNode, role string) (resource.Quantity, error) {
+	var q resource.Quantity
+	var err error
+	if role == MasterID {
+		q, err = resource.ParseQuantity(n.Memory)
+		if err != nil {
+			return resource.Quantity{}, microerror.Maskf(err, "creating Memory quantity from node definition")
+		}
+		additionalMemory, err := resource.ParseQuantity(baseMasterMemoryOverhead)
```
kopiczko: I would replace it with MustParse. If this value isn't parseable kvm-operator is basically unusable so it's fair to panic.

service/kvmconfig/v6/key/key.go
```diff
@@ -124,18 +128,54 @@ func MasterHostPathVolumeDir(clusterID string, vmNumber string) string {
...
+		if err != nil {
+			return resource.Quantity{}, microerror.Maskf(err, "creating Memory quantity from addtional memory")
+		}
+		q.Add(additionalMemory)
+
+	} else if role == WorkerID {
+		i, err := strconv.Atoi(n.Memory[:len(n.Memory)-1])
```
kopiczko: The code will panic when `n.Memory` is an empty string. So this will make whole operator fail because of wrong input.

Did you think about using https://godoc.org/k8s.io/apimachinery/pkg/api/resource#Quantity.Value ?
I think in case of memory value should be in bytes. Tests should confirm that.

service/kvmconfig/v6/key/key.go
```diff
@@ -124,18 +128,54 @@ func MasterHostPathVolumeDir(clusterID string, vmNumber string) string {
...
+		if err != nil {
+			return resource.Quantity{}, microerror.Maskf(err, "calculating memory overhead multiplier")
+		}
+		// base worker memory calculated in MB
+		baseMemoryMB := strconv.Itoa(i * 1024)
+
+		q, err = resource.ParseQuantity(baseMemoryMB + "M")
```
kopiczko: This could be just `q, err = resource.ParseQuantity(fmt.Sprintf("%dM", i * 1024))`. Or maybe even `q, err = resource.ParseQuantity(fmt.Sprintf("%dGi", i))`

service/kvmconfig/v6/key/key.go
```diff
@@ -124,18 +128,54 @@ func MasterHostPathVolumeDir(clusterID string, vmNumber string) string {
...
+		if err != nil {
+			return resource.Quantity{}, microerror.Maskf(err, "creating Memory quantity from node definition")
+		}
+		// IO overhead for qemu is around 512M memory
+		IOOverhead, err := resource.ParseQuantity(workerIOOverhead)
```
kopiczko: Same about `MustParse`

service/kvmconfig/v6/key/key.go
```diff
@@ -43,8 +43,12 @@ const (
 	K8SKVMHealthDocker        = "quay.io/giantswarm/k8s-kvm-health:ddf211dfed52086ade32ab8c45e44eb0273319ef"
 	NodeControllerDockerImage = "quay.io/giantswarm/kvm-operator-node-controller:7146561e54142d4f986daee0206336ebee3ceb18"
 
-	// defaultWorkerMemory represents the extra memory to add due to qemu overhead.
-	defaultWorkerMemory = "1G"
+	// constants for calculation qemu memory overhead.
+	baseMasterMemoryOverhead     = "1G"
+	baseWorkerMemoryOverhead     = 512
```
JosephSalisbury: I would specify the unit in the type, e.g: baseWorkerMemoryOverheadMb

service/kvmconfig/v6/key/key.go
calvix: splitted

service/kvmconfig/v6/key/key.go
calvix: done

service/kvmconfig/v6/key/key.go
calvix: done

service/kvmconfig/v6/key/key.go
calvix: done

service/kvmconfig/v6/key/key.go
calvix: done

calvix:
Can you guys take another look? thanks!

service/kvmconfig/v6/key/key.go
```diff
@@ -140,6 +143,41 @@ func MemoryQuantity(n v1alpha1.KVMConfigSpecKVMNode) (resource.Quantity, error)
...
+	}
+
+	// base worker memory calculated in MB
+	q, err := resource.ParseQuantity(fmt.Sprintf("%dM", mQuantity.ScaledValue(resource.Giga)*1024))
+	if err != nil {
+		return resource.Quantity{}, microerror.Maskf(err, "creating Memory quantity from node definition")
+	}
+	// IO overhead for qemu is around 512M memory
+	IOOverhead := resource.MustParse(workerIOOverhead)
+	q.Add(IOOverhead)
```
kopiczko: Nitpick: golang vairiables start with lowercase letter, so `ioOverhead`.

service/kvmconfig/v6/key/key.go
```diff
@@ -140,6 +143,41 @@ func MemoryQuantity(n v1alpha1.KVMConfigSpecKVMNode) (resource.Quantity, error)
...
+	mQuantity, err := resource.ParseQuantity(n.Memory)
+	if err != nil {
+		return resource.Quantity{}, microerror.Maskf(err, "calculating memory overhead multiplier")
+	}
+
+	// base worker memory calculated in MB
+	q, err := resource.ParseQuantity(fmt.Sprintf("%dM", mQuantity.ScaledValue(resource.Giga)*1024))
+	if err != nil {
+		return resource.Quantity{}, microerror.Maskf(err, "creating Memory quantity from node definition")
+	}
```
kopiczko: I don't understand why this can't just be following?

```go
	m, err := resource.ParseQuantity(n.Memory)
	if err != nil {
		return resource.Quantity{}, microerror.Maskf(err, "calculating memory overhead multiplier")
	}
```

Why this line is needed `q, err := resource.ParseQuantity(fmt.Sprintf("%dM", mQuantity.ScaledValue(resource.Giga)*1024))`?

calvix: because 8GB doesn't equal 8*1024

service/kvmconfig/v6/key/key.go
calvix: yeah i was thinking about that, IO is usually written with uppercase so wasnt sure, but its not really issue

service/kvmconfig/v6/key/key.go
kopiczko: But that's the value specified by the user. `8G == 8*1000`, `8Gi == 8*1024`.

calvix: the difference is important here

kopiczko: But this means we deal with `G` different way than kubernetes does.

calvix: we set the limits in `M` not `G` the limits just dont round up nicely with G

kopiczko: Whatever, then kvm-operator deal with `M` different way than kubernetes does.
Say at some point kvm-operator starts supporting `Mi` what we do then?

It isn't all that important, because we probably won't support anything else than `M`, but still confusing.

OK. I gave my LGTM. This is tested. I understand what's going on. Important thing is user doesn't know how it's computed and we can change it any time.

PR #412, opened 2018-02-27
add-annotation-for-prommy

calvix:
towards https://github.com/giantswarm/giantswarm/issues/2587 and PM https://github.com/giantswarm/giantswarm/issues/2645

Adding this annotation will  enable monitoring all the worker endpoints

### DIFF
--- a/service/kvmconfig/v6/resource/service/master_service.go
+++ b/service/kvmconfig/v6/resource/service/master_service.go
@@ -23,6 +23,10 @@ func newMasterService(customObject v1alpha1.KVMConfig) *apiv1.Service {
 			},
 			Annotations: map[string]string{
 				"giantswarm.io/prometheus-cluster": key.ClusterID(customObject),
+				"prometheus.io/path":               "/healthz",
+				"prometheus.io/port":               "'30010'",
+				"prometheus.io/scheme":             "http",
+				"prometheus.io/scrape":             "true",
 			},
 		},
 		Spec: apiv1.ServiceSpec{
--- a/service/kvmconfig/v6/resource/service/master_service.go
+++ b/service/kvmconfig/v6/resource/service/master_service.go
@@ -24,7 +24,7 @@ func newMasterService(customObject v1alpha1.KVMConfig) *apiv1.Service {
 			Annotations: map[string]string{
 				"giantswarm.io/prometheus-cluster": key.ClusterID(customObject),
 				"prometheus.io/path":               "/healthz",
-				"prometheus.io/port":               "'30010'",
+				"prometheus.io/port":               "30010",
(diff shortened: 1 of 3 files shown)


service/kvmconfig/v6/resource/service/master_service.go
```diff
@@ -23,6 +23,10 @@ func newMasterService(customObject v1alpha1.KVMConfig) *apiv1.Service {
 			},
 			Annotations: map[string]string{
 				"giantswarm.io/prometheus-cluster": key.ClusterID(customObject),
+				"prometheus.io/path":               "/healthz",
+				"prometheus.io/port":               "'30010'",
```
JosephSalisbury: Why the double quotes?

service/kvmconfig/v6/resource/service/worker_service.go
```diff
@@ -22,6 +22,12 @@ func newWorkerService(customObject v1alpha1.KVMConfig) *apiv1.Service {
 				"customer": key.ClusterCustomer(customObject),
 				"app":      key.WorkerID,
 			},
+			Annotations: map[string]string{
+				"prometheus.io/path":   "/healthz",
+				"prometheus.io/port":   "'30010'",
```
JosephSalisbury: Same here, we want just `"30010"`

calvix:
ping @JosephSalisbury  fixed in commit `single double quotes it is !`

MarcelMue:
Dumb question: why are we not using https here? Looks good otherwise.

JosephSalisbury:
Largely because we don't need to - this is pretty much an uptime check, we're not getting any metrics from the ingress controller.

calvix:
yeah we care if TCP is avaiable that is all

service/kvmconfig/v6/resource/service/master_service.go
```diff
@@ -23,6 +23,10 @@ func newMasterService(customObject v1alpha1.KVMConfig) *apiv1.Service {
 			},
 			Annotations: map[string]string{
 				"giantswarm.io/prometheus-cluster": key.ClusterID(customObject),
+				"prometheus.io/path":               "/healthz",
+				"prometheus.io/port":               "30010",
```
r7vme: 443 ???

r7vme: please ignore my comment. all good.

calvix: we don't wanna scrape api here, that is already taken care off by `prometheus-config-controller`
We discussed this with @JosephSalisbury and it's still beneficial for us to monitor master endpoint ip like this.

PR #429, opened 2018-03-07
Add WIP version 2.1.1

corest:


service/kvmconfig/v8/version_bundle.go
```diff
@@ -9,39 +9,9 @@ import (
...
-			},
-			{
-				Component:   "monitoring",
-				Description: "Added configuration for monitoring endpoint IP addresses.",
-				Kind:        versionbundle.KindAdded,
-			},
 			{
 				Component:   "cloudconfig",
-				Description: "Enable aggregation layer to be able to extend kubernetes API.",
+				Description: "Removed set-ownership-etcd-data-dir.service.",
```
kopiczko: Do you know this already changed?

corest: it will be there because of my next change, but I agree it is better to keep the history, so I'm updating this with placeholder

PR #468, opened 2018-04-19
Update to latest 3.2.5 k8scc

r7vme:
This changes are necessary to make iscsi in guest clusters to work properly.

Going to test this before merging.


r7vme:
tested in `geckon` merging

PR #474, opened 2018-04-27
Update certs dependency

tuommaki:
Update certs dependency to fix pressing issue on missing
calico-etcd-client cert on old clusters.


tuommaki:
Successfully tested on `gorgoth`. Gonna proceed once CI is happy.

PR #487, opened 2018-06-01
Change WIP version to minor

corest:
As my change with OIDC is a feature, there should be a minor version update


tuommaki (approved):
Just a reminder that this won't reflect to release version in release index as that version is determined by human.

PR #525, opened 2018-08-07
Expose ingress via nodePort

r7vme:
part-of: https://github.com/giantswarm/ingress-operator/issues/96

### Migration part

Make sure older cluster can be upgraded. TBD

### QA

- [x] new cluster has ingress available and exposed via nodePort
- [x] new cluster has no `ingressconfig` CR object
- [x] new cluster w/o [kubernetesd change](https://github.com/giantswarm/kubernetesd/pull/397) created w/o nodePort


service/controller/v14/key/key.go
```diff
@@ -341,3 +342,28 @@ func VersionBundleVersionFromPod(pod *corev1.Pod) (string, error) {
 func VMNumber(ID int) string {
 	return fmt.Sprintf("%d", ID)
 }
+
+func PortMappings(customObject v1alpha1.KVMConfig) []corev1.ServicePort {
+	var ports []corev1.ServicePort
+
+	//for _, p := range customObject.Spec.KVM.PortMappings {
+	//}
```
TheoBrigitte: Can we remove this ?

service/controller/v14/key/key.go
```diff
@@ -341,3 +342,28 @@ func VersionBundleVersionFromPod(pod *corev1.Pod) (string, error) {
...
+
+	// Return default port mapping
+	ports = []corev1.ServicePort{
+		{
+			Name:       "http",
+			Port:       int32(30010),
```
TheoBrigitte: Would be good to have ports number as constants IMO

PR #528, opened 2018-08-09
enable-real-e2e

calvix:
enable kvm e2e in gastropod, with new e2e harness


integration/teardown/teardown.go
```diff
@@ -3,39 +3,149 @@
...
+		})
+		if err != nil {
+			return microerror.Mask(err)
+		}
+		if len(certList.Items) == 0 {
+			// resource doesnt exist, we are good to continue
+			certDeleted = true
+			l.Log("level", "info", "message", "cert crd was deleted")
+		} else {
+			l.Log("level", "info", "message", "cert crd has not been deleted")
```
MarcelMue: You could just terminate early here with `return resourceNotDeleted` IMO. Then you won't need the variables and all three have to be deleted anyway, right?

Uncertain if this is necessary to change, just something I noticed.

MarcelMue (commented):
This is a lot of stuff. Tried to provide some feedback but it feels like I am missing parts of the picture.

integration/template/node_operator_charts_values.go
```diff
@@ -0,0 +1,11 @@
...
+
+const NodeOperatorChartValues = `
+clusterName: ${CLUSTER_NAME}
+Installation:
+  V1:    
+    Secret:      
+      Registry:
+        PullSecret:
+          DockerConfigJSON: "{\"auths\":{\"quay.io\":{\"auth\":\"$REGISTRY_PULL_SECRET\"}}}"
+`
```
xh3b4sd: Everything for the node operator should already be there and should then not be defined specifically in the `kvm-operator` repo. Further I think the `node-operator` is public and the general tone recently was to get rid of pull secrets for public services. Uncertain though.

integration/utils/utils.go
```diff
@@ -0,0 +1,217 @@
...
+
+import (
+	"context"
+	"fmt"
+	"net"
+
+	"k8s.io/api/core/v1"
+	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
```
xh3b4sd: All these imports should be in the same block.

integration/teardown/teardown.go
```diff
@@ -3,39 +3,149 @@
...
 
 import (
+	"context"
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/giantswarm/backoff"
 	"github.com/giantswarm/e2e-harness/pkg/framework"
+	"github.com/giantswarm/kvm-operator/integration/utils"
```
xh3b4sd: Please see https://github.com/giantswarm/fmt/blob/master/go/imports.md.

integration/utils/utils.go
```diff
@@ -0,0 +1,217 @@
...
+	if len(items) != 1 {
+		return 0, microerror.Maskf(executionFailedError, "range pool VNI generation failed, expected 1 got %d", len(items))
+	}
+
+	return items[0], nil
+}
+
+func DeleteVNI(rangePool *rangepool.Service, clusterID string) error {
+	return rangePool.Delete(context.Background(), vniRangepoolNamespace, rangePoolVNIID(clusterID))
```
xh3b4sd: Lets please always make the error handling explicit so the story the stack tells you in error cases reflects reality.

xh3b4sd: I think it would be good to propagate a context here. We can create one when the actual test starts and dispatch it through. So we do not have to create one on each new level, because this is rather useless.

integration/utils/utils.go
```diff
@@ -0,0 +1,217 @@
...
+
+	cidr, err := ipamService.CreateSubnet(context.Background(), cidrMask, flannelNetworkAnnotation(clusterID))
+	if err != nil {
+		return "", microerror.Mask(err)
+	}
+
+	return cidr.String(), nil
+}
+
+func DeleteFlannelNetwork(network string, crdStorage microstorage.Storage, l micrologger.Logger) error {
```
xh3b4sd: For @giantswarm/sig-operator all this looks like the `kvm-operator` is less self contained than the other provider operators. Something to consider in the long run that we should try to improve the design and make responsibilities more clear. Right now they are quite scattered.

xh3b4sd (changes requested):

integration/setup/setup.go
```diff
@@ -83,6 +95,152 @@ func Resources(g *framework.Guest, h *framework.Host) error {
...
+			return microerror.Mask(err)
+		}
+		flannelResourceChartValues.Network = network
+	}
+
+	o := func() error {
+		// NOTE we ignore errors here because we cannot get really useful error
+		// handling done. This here should anyway only be a quick fix until we use
+		// the helm client lib. Then error handling will be better.
+		framework.HelmCmd(fmt.Sprintf("delete --purge %s-flannel-config-e2e", h.TargetNamespace()))
```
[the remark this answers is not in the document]
calvix: any example of how we use that?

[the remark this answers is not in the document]
MarcelMue: I just realised that we don't have a nice function / example of doing this yet. Ignore this comment for now, we can improve in the future here.

integration/teardown/teardown.go
calvix: Good idea, will fix it.

integration/setup/setup.go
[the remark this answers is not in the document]
calvix: ack

integration/utils/utils.go
calvix: fixed

integration/teardown/teardown.go
calvix: fixed

integration/env/common.go
```diff
@@ -137,10 +140,19 @@ func ClusterID() string {
 
 	parts = append(parts, "ci")
 	parts = append(parts, TestedVersion()[0:3])
-	parts = append(parts, CircleSHA()[0:5])
-	if TestHash() != "" {
-		parts = append(parts, TestHash())
-	}
+	shaPart := CircleSHA()[0:4]
+	testPart := TestHash()
```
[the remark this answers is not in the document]
calvix: AFAIK TestHash is hash of integration test directory, so we have some unique identificator across parallel tests.

The magic is there because i need to somehow reduce the number of letters in clusterID so I do hash from the 2 hashes.

integration/env/common.go
```diff
@@ -137,10 +140,19 @@ func ClusterID() string {
...
+
+	h := sha1.New()
+	h.Write([]byte(shaPart + testPart))
+	s := fmt.Sprintf("%x", h.Sum(nil))[0:5]
+
+	parts = append(parts, s)
+	/*
```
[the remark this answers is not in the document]
calvix: fixed

kopiczko (approved):
The setup here is crazy. But the work you did is very nice in general. LGTM

kopiczko:
Maybe we should just move allocation responsibility from kubernetesd to kvm-operator (and other operators in general). Instead of preallocating stuff we can allocate it inside the operator and write allocated values in the status. Should we have issue for that?

calvix:
> Maybe we should just move allocation responsibility from kubernetesd to kvm-operator (and other operators in general). Instead of preallocating stuff we can allocate it inside the operator and write allocated values in the status. Should we have issue for that?

I think the reason why we have it in kubernetesd is that some values are needed in more CRDs/operators, but maybe there is a way how to solve that.

kopiczko:
@calvix you are right, it needs flannel to function. Solution would be:

- make flannel-operator a flannel controller in kvm-operator
- make flannel controller watch KVMConfigs
- make flannel controller do the allocations
- use initializers to prevent cluster/drainer controllers from running before flannel is initialized

integration/teardown/teardown.go
```diff
@@ -3,40 +3,143 @@
...
+	"fmt"
+
+	"github.com/giantswarm/backoff"
 	"github.com/giantswarm/e2e-harness/pkg/framework"
 	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/micrologger"
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
 
-	"github.com/giantswarm/kvm-operator/integration/env"
+	"github.com/giantswarm/kvm-operator/integration/utils"
```
xh3b4sd: FYI it is a kind of anti pattern in go to call packages `util` or something abstract like that by convention. We should try to be more specific. Such names are a sign of weak design.

integration/teardown/teardown.go
```diff
@@ -3,40 +3,143 @@
...
+	o := func() error {
+		certList, err := h.G8sClient().CoreV1alpha1().CertConfigs(v1.NamespaceDefault).List(v1.ListOptions{
+			LabelSelector: crdLabelSelector(clusterID),
+		})
+		if err != nil {
+			return microerror.Mask(err)
+		}
+		if len(certList.Items) == 0 {
+			// resource doesnt exist, we are good to continue
+			l.Log("level", "info", "message", "cert crd was deleted")
```
xh3b4sd: We seem to have a context available, so all logging should make use of it via `LogCtx`.

xh3b4sd (approved):
It is really impressive that you lifted that rock up the mountain. Thank you for all your efforts here. I am looking forward to extend KVM with more of our common tests.

integration/teardown/teardown.go
calvix: So separating the file into two packages `rangepool` and `ipam` would make more sense in this case?

xh3b4sd: When this is what the package essentially is about, then yes. It would at least be better than `utils`.

integration/teardown/teardown.go
calvix: fixed

calvix:
I need some help or advice here.
In the teardown phase there is this code:
https://github.com/giantswarm/kvm-operator/blob/e5d60cbf45c3c8a2b30434f81afaf779b289404f/integration/teardown/teardown.go#L43-L56

The problem I see is if the test goes wrong in setup phase (some error on network, someone cancel test or something else) and  some of the  CRDs are not created then these commands will fail to delete the CRD charts which will result in exit of the test and the rest of stuff is not cleaned properly.

Ignoring errors is in general bad idea but I don't see how could I do it here without ignoring the errors.
Any thoughts?

calvix:
> @kopiczko did something in AWS or Azure IIRC. Check the teardown there. And before you yell at me again, here is the link: https://github.com/giantswarm/aws-operator/blob/1fc2ae9aa574018eee5e0c1da4e8da2a295d0142/integration/setup/teardown.go#L17-L57. xD

Thanks! that's good

calvix:
OK, so everything should be now in place. This is the final code I wanted to have there.

PTAL and I will merge after.

xh3b4sd (approved):
When this works lets go and iterate. Awesome progress here. Thanks man.

MarcelMue (approved):
<3

PR #534, opened 2018-08-27
Disable Ingress Controller resources in k8scloudconfig

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/3710


rossf7:
Tested on `geckon`.

MarcelMue (approved):
I am not sure if and how the correct order of deployments is important here between `cluster-operator` and this. The change seems fine though.

PR #560, opened 2018-09-20
fix e2e-cluster-id

calvix:
fix broken cluster name
see https://github.com/giantswarm/apiextensions/blob/master/helm/apiextensions-kvm-config-e2e-chart/values.yaml -> https://github.com/giantswarm/apiextensions/pull/135


MarcelMue (approved):
Makes sense.

calvix:
hmm seems like this si already implemented here: https://github.com/giantswarm/kvm-operator/pull/555but its not merged

PR #568, opened 2018-10-02
Update statusresource dep

tuommaki:
Statusresource dependency has a fix to ensure default patches when
status might be completely missing.


tuommaki:
Tested on `gorgoth`. Proceeding here as this just updates dep.

PR #569, opened 2018-10-03
Retry operations that involve crdstorage in e2e tests

tuommaki:
On KVM e2e tests run on same control plane and in order to coordinate
Flannel VNI, Ingress node port and Flannel network allocations, all e2e
tests must use same crdstorage instance. This introduces race condition
where two e2e tests are running concurrently and modify the shared CR
object. To work with this, operations that involve crdstorage
modifications must retry on error.

Towards https://github.com/giantswarm/giantswarm/issues/4214

### DIFF
--- a/integration/ipam/ipam.go
+++ b/integration/ipam/ipam.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"net"
 
+	"github.com/giantswarm/backoff"
+	"github.com/giantswarm/e2e-harness/pkg/framework"
 	"github.com/giantswarm/ipam"
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/micrologger"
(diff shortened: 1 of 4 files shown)


integration/storage/storage.go
```diff
@@ -44,10 +44,15 @@ func InitCRDStorage(ctx context.Context, h *framework.Host, l micrologger.Logger
...
+	targetNamespace := h.TargetNamespace()
+	if targetNamespace == "" {
+		targetNamespace = giantswarmNamespace
+	}
+
 	c.Name = "kvm-e2e"
 	c.Namespace = &v1.Namespace{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: giantswarmNamespace,
+			Name: targetNamespace,
```
calvix: this is wrong, the storage has to be in  same namespace, if you separate storage by namesapce then each test will have its own storage and the whole rangepool magic wont work

tuommaki: Shit. You are totally right. We already discussed this. I reused the same branch so I must have committed it before cleaning it out (I did `git checkout..` for changes before working on retry).... Oh well, I'll remove this immediately. Thanks!

calvix (approved):

integration/ipam/ipam.go
```diff
@@ -37,7 +39,20 @@ func GenerateFlannelNetwork(ctx context.Context, clusterID string, crdStorage mi
 
 	cidrMask := net.CIDRMask(flannelCidrSize, 32)
```
kopiczko: Maybe that's helpful: https://github.com/giantswarm/microstorage/tree/master/retrystorage

tuommaki: Oh, that would be cleaner approach. I'll refactor this in the morning.

tuommaki:
Dropping this in favor of https://github.com/giantswarm/kvm-operator/pull/570

PR #570, opened 2018-10-04
Wrap crdstorage with retrystorage in e2e tests

tuommaki:
On KVM e2e tests run on same control plane and in order to coordinate
Flannel VNI, Ingress node port and Flannel network allocations, all e2e
tests must use same crdstorage instance. This introduces race condition
where two e2e tests are running concurrently and modify the shared CR
object. To work with this, operations that involve crdstorage
modifications must retry on error.

Towards https://github.com/giantswarm/giantswarm/issues/4214

PR #571, opened 2018-10-05
Vendor e2e-harness and use time const from backoff

indyfree:


indyfree:
As mentioned here: https://github.com/giantswarm/kvm-operator/pull/570#issuecomment-427268876 e2etest are still failing for a reason.

My vendoring change is unrelated. So I'd go ahead.

MarcelMue (approved):
LGTM

PR #573, opened 2018-10-08
Make clusterrole configurable

MarcelMue:
towards https://github.com/giantswarm/giantswarm/issues/4214

This makes it possible to configure the bindings, that way we can create different ones for each operator while we run kvm e2e concurrently. We hopefully won't see any more warnings about bindings existing anymore.

A PR for e2etemplates is also prepared already: https://github.com/giantswarm/e2etemplates/pull/27


helm/kvm-operator-chart/values.yaml
```diff
@@ -1 +1,5 @@
+clusterRoleBindingName: kvm-operator
```
MarcelMue: Feel free to make suggestions here, naming is hard.

xh3b4sd (approved):

calvix (approved):
looks good, needs to be tested  if it doesnt break stuff

MarcelMue:
Deployed this branch to `geckon` and everything is fine. I can see the operator reconciling and don't see any errors, so I assume everything is fine.

PR #583, opened 2018-10-12
Fix typo IsPodDraind -> IsPodDrained

tuommaki:
Drive-by typo fix while changing the function behavior.


MarcelMue (approved):
LGTM

tuommaki:
E2e is b0rked, but as this is only renaming of a function and build + unit tests are fine, I'm going with this.

PR #585, opened 2018-10-12
Use k8scc 3.6.2 for WIP release

r7vme:
Use latest 3.6.2 k8scc release for WIP version bundle of kvm-operator

### DIFF
--- a/service/controller/v16/version_bundle.go
+++ b/service/controller/v16/version_bundle.go
@@ -8,6 +8,26 @@ func VersionBundle() versionbundle.Bundle {
 	return versionbundle.Bundle{
 		Changelogs: []versionbundle.Changelog{
 			{
+				Component:   "cloudconfig",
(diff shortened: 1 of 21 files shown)


service/controller/v16/version_bundle.go
```diff
@@ -16,7 +36,7 @@ func VersionBundle() versionbundle.Bundle {
 		Components: []versionbundle.Component{
 			{
 				Name:    "calico",
-				Version: "3.0.5",
+				Version: "3.2.3",
```
r7vme: Not sure how, we should keep this up to date.

kopiczko: That can be the part of releasing operators story in our team.

kopiczko (approved):

service/controller/v16/version_bundle.go
kopiczko: Go ahead and add a point.

PR #588, opened 2018-10-12
Vendor fix from cert-operator templates

MarcelMue:


MarcelMue:
Note:
CI went green but the update test seems to flap. I will continue to try and find a root cause of that flap, but the operator logs don't show anything helpful so far.

xh3b4sd (approved):

PR #596, opened 2018-10-18
Fix node draining status check

tuommaki:
When draining tenant node, in case all containers are terminated within
POD but node is not drained yet according to DrainerConfig, consider it
as drained. Check for this was in wrong place and is now moved to right
place.


xh3b4sd (approved):
That should be better. Lets see what e2e is saying.

tuommaki:
All tests green. Merging.

PR #604, opened 2018-11-06
Disable CoreDNS k8scloudconfig templates

TheoBrigitte:
Towards https://github.com/giantswarm/giantswarm/issues/1902

Disable management of CoreDNS component via k8scloudconfig as it is now managed by cluster-operator.

### DIFF
--- a/service/controller/v16/version_bundle.go
+++ b/service/controller/v16/version_bundle.go
@@ -8,11 +8,6 @@ func VersionBundle() versionbundle.Bundle {
 	return versionbundle.Bundle{
 		Changelogs: []versionbundle.Changelog{
 			{
-				Component:   "cloudconfig",
-				Description: "Removed CoreDNS related components (now managed by chart-operator).",
-				Kind:        versionbundle.KindRemoved,
-			},
-			{
 				Component:   "kubernetes",
 				Description: "Updated Kubernetes to 1.12.2. More info here: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.12.md",
 				Kind:        versionbundle.KindChanged,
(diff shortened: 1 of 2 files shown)


service/controller/v16/version_bundle.go
```diff
@@ -7,6 +7,11 @@ import (
 func VersionBundle() versionbundle.Bundle {
 	return versionbundle.Bundle{
 		Changelogs: []versionbundle.Changelog{
+			{
```
teemow: i would not mention this here. we have a changelog in cluster-operator about coredns being configurable by the user. this sounds like it got removed.

teemow (approved):
LGTM besides the changelog

service/controller/v16/version_bundle.go
teemow: did we have those changelog messages in aws and azure in older releases? because in the k8s 1.12 release those also switched to cluster-operator 0.8 and they didn't have this message. 

my thoughts were that this could be confusing. for the user it is not clear. is it removed or configurable now? :smile: 

that why i thought to keep it similar to the other operators and keep the cluster-operator changelog.

btw release notes for kvm are here and you can compare them with aws:

* KVM: https://github.com/giantswarm/giantswarm/issues/4312#issuecomment-435999048
* AWS: https://github.com/giantswarm/giantswarm/issues/4312#issuecomment-436040368

teemow: I don't care to have consistency between operators. More importantly we should have a good and clear message in the release.

So the result will be:

 * cloudconfig: Removed CoreDNS related components (now managed by chart-operator).
 * coredns: Allow advanced CoreDNS configuration. See https://docs.giantswarm.io/guides/advanced-coredns-configuration
or
 * coredns: Add coredns chartconfig.

I would be fine with just having the one with the link. Or explain the others more.

TheoBrigitte: ok, removed the changelog message

PR #605, opened 2018-11-07
update kvm sidecars with proper signal handling

calvix:
* update k8s-endpoint-updater image + remove execution from bash command
* update k8s-kvm-health image with the changes for microkit signal handling
* update shutdown-deferrer image with changes for microkit signal handling


service/controller/v15/resource/deployment/master_deployment.go
```diff
@@ -165,13 +165,12 @@ func newMasterDeployments(customObject v1alpha1.KVMConfig) ([]*extensionsv1.Depl
 								Image:           key.K8SEndpointUpdaterDocker,
 								ImagePullPolicy: apiv1.PullIfNotPresent,
 								Command: []string{
-									"/bin/sh",
-									"-c",
-									"/opt/k8s-endpoint-updater update --provider.bridge.name=" + key.NetworkBridgeName(customObject) +
```
teemow: you could also just run `exec /opt/k8s-endpoint-updater`. then you wouldn't have to change the flag in the endpoint-updater.

calvix: would exec also interpret the env value?

teemow: the env value should be fine. if signal handling works fine :man_shrugging:. was just an idea.

tuommaki: When overriding `ENTRYPOINT` and `CMD` (that are defined in docker container), isn't it then executing given command as such and then it must be full path to executable in container. There can't be any shell in between because that would be implicit requirement for containers and I'm not aware of anything like this. Or am I missing something?

calvix: Yes it is like that, so  looks like exec is not usable

calvix:
tested and this works nicely no more hanging fruit
sidecars are stopped right away, deferrer waits for  draining and then k8s-kvm is terminted via qemu shutdown

3 worker cluster on gorgoth ( no workloads) was upgraded  from 3.0.0 to 4.0.0 in 15m

I still wanna sted the upgrade process a few times to use how reliable it is.
My plan is to merge tomorrow morning.

teemow:
thanks for the heads up. my tests were also successful. so we can craft the 4.0.0 release tomorrow morning.

calvix:
another test went smoothly, merging

PR #631, opened 2018-12-17
vendor: updated helmclient and removed service account files

rossf7:
Towards giantswarm/giantswarm#4352


rossf7:
Closing. Changes already vendored.

PR #643, opened 2019-01-31
add-static-initiator-name-for-vms

calvix:
the initiator name is built from  cluster-id, node id and node role ie: 
```
iqn.2016-04.com.coreos.iscsi:giantswarm-${CLUSTER_ID}-${NODE_ROLE}-${NODE_ID}
iqn.2016-04.com.coreos.iscsi:giantswarm-xsd4e-master-w23kl
```

this solution should help a customer with using iscsi on KVM clusters as the initiators will be always the same for the same node and easily deterministic

The primary source for this change is the current centaur situation and their need to somehow resolve the story and if more customer wants to try iscsi this will help us in the future. 

There should be no drawback.

### DIFF
--- a/service/controller/v18/version_bundle.go
+++ b/service/controller/v18/version_bundle.go
@@ -27,6 +27,11 @@ func VersionBundle() versionbundle.Bundle {
 				Description: "Improved node draining during updates and scaling.",
 				Kind:        versionbundle.KindChanged,
 			},
+			{
+				Component:   "cloudconfig",
+				Description: "Add static initiator name for each vm.",
+				Kind:        versionbundle.KindAdded,
+			},
 		},
 		Components: []versionbundle.Component{
 			{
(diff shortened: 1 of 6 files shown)


tuommaki (approved):
LGTM when it works for you :+1:

calvix:
adding also config file to workaround this bug https://github.com/kubernetes/kubernetes/issues/73181

service/controller/v18/version_bundle.go
```diff
@@ -27,6 +27,11 @@ func VersionBundle() versionbundle.Bundle {
 				Description: "Improved node draining during updates and scaling.",
 				Kind:        versionbundle.KindChanged,
 			},
+			{
+				Component:   "cloudconfig",
+				Description: "Add static initiator name for each vm.",
```
kopiczko: Should it be that?

```suggestion
				Description: "Add static iSCSI initiator name for each vm.",
```

PR #644, opened 2019-02-07
Vendor update statusresource

tuommaki:
Update latest statusresource in order to get node labels into node list in
KVMConfig.Status for older clusters as well (without cluster scaling/update).


MarcelMue (approved):

tuommaki:
AWS & Azure are cool. KVM e2e not work. This going with me it is.

PR #652, opened 2019-02-14
use-node-index-instead-node-id

calvix:
to made initiatorName more predictable  switching to node index instead of nodeID
Because if the customer wants to scale nodes he needs to allow initiatorNames before or otherwise, he risks that on new nodes storage won't work. 

this goes from centaur requirements and kinda make sense


calvix:
> Even though we don't do that this is perfectly fine to change order of nodes in the kvmconfig spec. If someone has fantasy to do it or we e.g. write tooling to change something and don't pay attention we can run into issues. I have mixed feelings here.

That's true, but I think we already rely on a few things already, like fi someone change name of something in CRD we would be screw too.

Anyway, if the order change and someone roll update than two nodes will have swapped initiator names but that won't break anything. The only issue would be if someone changed stuff and then rolled only one worker.

Do you have any other idea how we can do this so it will be predictable?


If not we will have to make an announcement that the order of worker should not be meddled with.

kopiczko:
The only reliable way I see is to maintain this information in CR status and do proper allocations.

calvix:
> The only reliable way I see is to maintain this information in CR status and do proper allocations.

can you explain a little bit further how I can achieve this?

calvix:
ping @tuommaki

tuommaki:
Intermediate information here: There's now open PR for maintaining node index in status: https://github.com/giantswarm/kvm-operator/pull/661

kopiczko:
What @tuommaki proposed in #661 is exactly what I meant.

calvix (commented):
cant approve as its my own PR, but LGTM

kopiczko (approved):

PR #660, opened 2019-02-25
vendor k8scloudconfig: switch kube-proxy back to iptables

teemow:
ipvs is still experimental and with 1.12.x clusters we are still having
issues with it. so for now we go back to 'iptables'


teemow:
I just did a `dep ensure -update github.com/giantswarm/k8scloudconfig` but there were more updates than in the other operators. so uncertain what else changed. I'll give this another shot on a test installation.

teemow:
Ok this is borked anyway `cannot find package "github.com/aokoli/goutils"`

PR #674, opened 2019-03-05
Backport iSCSI initiator name fix from v19

tuommaki:
This change backports a fix for iSCSI initiator name that makes it use assigned
node index instead of node ID in order to be predictable and unique among
running nodes.

Original PR: https://github.com/giantswarm/kvm-operator/pull/652


corest (approved):
lgtm

PR #676, opened 2019-03-05
Update CoreOS

mazzy89:


service/controller/v19/version_bundle.go
```diff
@@ -12,6 +12,11 @@ func VersionBundle() versionbundle.Bundle {
 				Description: "Changed iSCSI initiator name to use assigned node index instead of node ID.",
 				Kind:        versionbundle.KindChanged,
 			},
+			{
+				Component:   "containerlinux",
+				Description: "Update to 2023.4.0.",
```
teemow:
```suggestion
				Description: "Update to 2023.4.0. Fixes CVE-2019-8912",
```

mazzy89: Shouldn't be better to put the changelog regarding the CVE in a different block which has as component `kubernetes`?

teemow: It is a linux kernel cve that was fixed in this container linux version.

PR #679, opened 2019-03-05
Upgrades kubernetes CVE-2019-1002100

ferrandinand:
~~- Creates v17patch2~~ 
- Updates v18patch1
- Updates v19

Towards: giantswarm/giantswarm#5506


service/controller/v17patch2/version_bundle.go
```diff
@@ -8,8 +8,8 @@ func VersionBundle() versionbundle.Bundle {
 	return versionbundle.Bundle{
 		Changelogs: []versionbundle.Changelog{
 			{
-				Component:   "containerlinux",
-				Description: "Update to 1967.5.0 (CVE-2019-5736).",
+				Component:   "kubernetes",
+				Description: "Update to 1.12.6 (CVE-2019-1002100).",
 				Kind:        versionbundle.KindChanged,
```
calvix: you want to add new changelog entry instead if editing already existing

ferrandinand: This comes from the full copy of process of v17patch1.
So that entry belongs to v17patch1 and I guess it is not required in v17patch2. Isn´t it?

rossf7 (commented):
There are a couple of wiring changes needed.

- New version bundles need to be added here.
https://github.com/giantswarm/kvm-operator/blob/master/service/version_bundles.go

- You also need to add the new resource sets here.
https://github.com/giantswarm/kvm-operator/blob/106ca62a9f67573aae7504abfd878bd3b97a8c9f/service/controller/cluster.go#L378

Also not 100% but looks like the k8scloudconfig changes are not vendored yet.

teemow:
@ferrandinand as mentioned in the other issue. We probably don't need a patch for 1.12 on KVM. 

https://github.com/giantswarm/giantswarm/issues/5506#issuecomment-469702511
https://github.com/giantswarm/giantswarm/issues/5506#issuecomment-469631444

ferrandinand:
@teemow Ok, I see this now. So, that makes this simpler. 
Let´s remove v17patch2 version then.

service/controller/v17patch2/version_bundle.go
calvix: right,  i thought that you are rewriting the new coreos

ferrandinand:
> Also not 100% but looks like the k8scloudconfig changes are not vendored yet.

I thought It was better to not include this PR for make it easier to review. Where you expecting all vendoring stuff in this PR?

rossf7:
@ferrandinand Problem is the build step is failing because it needs the k8scc update to compile.

To avoid a huge vendor update you can just update that dependency.

`dep ensure -update github.com/giantswarm/k8scloudconfigdep`

ferrandinand:
@rossf7 great I´m going to add the vendor files here then

rossf7 (approved):
LGTM when tested.

PR #682, opened 2019-03-12
import kvm health  fix for closing idle connections

calvix:
importing https://github.com/giantswarm/k8s-kvm-health/pull/9


calvix:
tested on geckon and anubis, all good
even update was green at the end

merging

PR #687, opened 2019-03-19
increase-master-mem-overhead

calvix:
towards https://github.com/giantswarm/giantswarm/issues/5547

on  big KVM clusters, master gets more IO and more load which cause reaching the memory limit, so adjusting here to also include memory  overhead for IO

### DIFF
--- a/service/controller/v20/key/key.go
+++ b/service/controller/v20/key/key.go
@@ -63,7 +63,7 @@ const (
 	baseWorkerMemoryOverheadMB   = 512
 	baseWorkerOverheadMultiplier = 2
 	baseWorkerOverheadModulator  = 12
-	workerIOOverhead             = "512M"
+	qemuMemoryIOOverhead         = "512M"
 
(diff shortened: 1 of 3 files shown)


service/controller/v20/key/key.go
```diff
@@ -318,11 +318,12 @@ func MemoryQuantityMaster(n v1alpha1.KVMConfigSpecKVMNode) (resource.Quantity, e
 		return resource.Quantity{}, microerror.Maskf(err, "creating Memory quantity from node definition")
 	}
 	additionalMemory := resource.MustParse(baseMasterMemoryOverhead)
-	if err != nil {
-		return resource.Quantity{}, microerror.Maskf(err, "creating Memory quantity from addtional memory")
-	}
```
calvix: this was useless

kopiczko (approved):
LGTM if we can fit existing clusters.

calvix:
checked all KVM clusters and masters still have lot of free resources so we are fine

PR #699, opened 2019-04-15
Fix kvm-operator rbac

corest:
Towards https://github.com/giantswarm/giantswarm/issues/1912


rossf7 (commented):
Looks fine but why are these changes needed? Can you link an issue?

corest:
@rossf7 linked an issue. RBAC onprem was blocked because of flannel. It is not blocked anymore now, so I've enabled strict RBAC on test installation and looking into rules which should be updated

rossf7 (approved):
Thanks! LGTM

MarcelMue (approved):

PR #712, opened 2019-06-05
improve-memory-overhead-formulat-with-new-qemu

calvix:
improve memory overhead formula to  work with new qemu version

 after checking anubis cluster, it looks like all clusters are experiencing the same issue so its across whole range of memory usage

for https://github.com/giantswarm/giantswarm/issues/6120


calvix:
this is just educated guess based on metrics available, it's super hard to test it properly as we see the issue only for one customer and customer often run workers with 80GB+ RAM which we can't in our environment

corest (approved):
lgtmiwfy

tuommaki (approved):
When this works :+1:

PR #720, opened 2019-06-25
run as non-root user

glitchcrab:
Towards giantswarm/giantswarm/issues/6013


glitchcrab:
can't quite figure out what is causing this one to go bang:

```
panic: Post https://g8s.gastropod.gridscale.kvm.gigantic.io:443/apis/apiextensions.k8s.io/v1beta1/customresourcedefinitions: EOF
```

glitchcrab:
Tested and working fine on geckon.

helm/kvm-operator-chart/templates/psp.yaml
```diff
@@ -5,16 +5,24 @@ metadata:
...
+    rule: MustRunAs
+    ranges:
+      - min: 1
+        max: 65535
   seLinux:
     rule: RunAsAny
   supplementalGroups:
     rule: RunAsAny
   volumes:
-    - 'secret'
```
rossf7: For my interest why don't we need secret? Is there no volume?

glitchcrab: Yup that's correct, only the configmap is mounted as a volume: https://github.com/giantswarm/kvm-operator/blob/master/helm/kvm-operator-chart/templates/deployment.yaml#L20

rossf7: Cool thanks for the explanation and good to clean up. 👍

PR #735, opened 2019-07-15
vendor: updated k8scloudconfig

rossf7:
Towards giantswarm/giantswarm#6367


rossf7:
Vendor PR going with a ping.

PR #748, opened 2019-08-12
Update kubernetes 1.14.5 (CVE-2019-11247)

cornelius-keller:


service/controller/v23patch1/version_bundle.go
```diff
@@ -8,8 +8,8 @@ func VersionBundle() versionbundle.Bundle {
 	return versionbundle.Bundle{
 		Changelogs: []versionbundle.Changelog{
 			{
-				Component:   "TODO",
-				Description: "TODO",
+				Component:   "kubernetes",
+				Description: "Update kubernetes to 1.14.5 (CVE-2019-11247)",
```
MarcelMue: We should put a link into this description, see here for example: https://github.com/giantswarm/kvm-operator/blob/master/service/controller/v23/version_bundle.go#L12

schjan:
I don't know how to comment unchanged files but you miss one place to update the versionbundle at https://github.com/giantswarm/kvm-operator/blob/kubernetes-1.14.5/service/controller/v23patch1/cloudconfig/cloudconfigtest/cloudconfigtest.go

service/controller/v23patch1/version_bundle.go
cornelius-keller: fixed

cornelius-keller:
> I don't know how to comment unchanged files but you miss one place to update the versionbundle at https://github.com/giantswarm/kvm-operator/blob/kubernetes-1.14.5/service/controller/v23patch1/cloudconfig/cloudconfigtest/cloudconfigtest.go
It seems somebody forgot this already the last time. Did not see it because it was still on `v_4_4_0`

PR #770, opened 2019-11-04
Add resource version to endpoint calculation

MarcelMue:
This adds the resource version to the calculation and therefor ensures that the endpoint we update is in the same state as the one we read information from.

During a hangout with Tim we discovered that the following situation is likely causing us problems:
- We start the update of a cluster
- The old pods have a label with the old `VersionBundleVersion`
- Each new pod that gets started by the rolling update has the new `VersionBundleVersion`
- All pods with the old `VBV` are reconciled by the old controller (e.g. V1)
- All pods with the new `VBV` are reconciled by the new controller (e.g. V2)
- Both controllers are now editing the _single_ service which contains all endpoints for old and new pods
- Since neither controller is checking the resource version we get the following situation:
    - V1 reads the endpoints (E1,E2,E3)
    - V2 reads the endpoints (E1,E2,E3)
    - V1 adds the new endpoint E4 (E1,E2,E3,E4)
    - V2 removes an old endpoint E1 (E2,E3)
    - V2 writes its changes to k8s -> (E2,E3)
    - V1 does not check the resourceVersion and writes its changes to k8s -> (E1,E2,E3,E4)
-> We end up with one endpoint (E1) which has no more pod and both controllers exited without error 

towards https://github.com/giantswarm/giantswarm/issues/7283

### DIFF
--- a/service/controller/v26/resource/endpoint/current.go
+++ b/service/controller/v26/resource/endpoint/current.go
@@ -72,6 +72,13 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 			return nil, microerror.Mask(err)
 		}
 
+		if k8sEndpoints != nil {
+			// If we found a k8s endpoint then we set this resource version here.
+			// We do this to later guarantee that the endpoint did not change
+			// between gathering information and submitting our update.
+			endpoint.ResourceVersion = k8sEndpoints.ResourceVersion
+		}
+
 		for _, endpointSubset := range k8sEndpoints.Subsets {
 			for _, endpointAddress := range endpointSubset.Addresses {
 				if !containsIP(endpoint.IPs, endpointAddress.IP) {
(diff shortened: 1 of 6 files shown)


MarcelMue:
Need to test still but reviews / ideas welcome.

xh3b4sd (approved):
Minimal effort, maximum outcome.

tuommaki (approved):
Makes sense

service/controller/v26/resource/endpoint/current.go
```diff
@@ -72,6 +72,13 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 			return nil, microerror.Mask(err)
 		}
 
+		if k8sEndpoints != nil {
```
kopiczko: Can `k8sEndpoints` be `nil` when `errors.IsNotFound` is handled with `return`?

MarcelMue: You are correct, I was a bit too paranoid because we allow the service to be nil.

PR #774, opened 2019-11-15
Test systemd kubelet

tfussell:


tfussell:
This was only used for testing. It will reappear for testing the full 1.16 release. Closing now.

PR #775, opened 2019-11-19
Drop calico-kube-kill.service

tfussell:
While working on non-containerized kubelet, I noticed that this service was failing silently because it used `kubectl` v1.10 and `kubectl get cs` didn't work correctly due to the version mismatch. I updated `docker-kubectl` to v1.15, updated the kubeconfig location, and removed coredns since it's an app instead of a core component now. I'm curious why this wasn't noticed before. Maybe it's not necessary at all now or only for certain installations?


mazzy89:
Nice catch. in the last kvm clusters I've spinned up, coredns was installed via app-operator.

mazzy89 (approved):

JosephSalisbury:
Do we have something tracking why we weren't alerted to this failing? Do we even need it, if it can fail silently and no-one is worried?

MarcelMue:
It seems to me like the silent failure is the real problem here. 
Why was the unit reporting as a failed unit in systemd?

If we actually need / use the unit is a question I cannot answer - I do know that we alert on failed units.

service/controller/v26/cloudconfig/master_template.go
```diff
@@ -225,7 +225,6 @@ ExecStart=/bin/sh -c '\
 		sleep 1m && \
 		/usr/bin/docker run -e KUBECONFIG=${KUBECONFIG} --net=host --rm -v /etc/kubernetes:/etc/kubernetes $KUBECTL -n kube-system delete pod -l k8s-app=kube-proxy && \
 		/usr/bin/docker run -e KUBECONFIG=${KUBECONFIG} --net=host --rm -v /etc/kubernetes:/etc/kubernetes $KUBECTL -n kube-system delete pod -l k8s-app=calico-kube-controllers && \
-		/usr/bin/docker run -e KUBECONFIG=${KUBECONFIG} --net=host --rm -v /etc/kubernetes:/etc/kubernetes $KUBECTL -n kube-system delete pod -l k8s-app=coredns &&\
 		result="ok" || echo "failed";\
 	done;\
 	[ "$result" != "ok" ] && echo "Failed to restart k8s services." && exit 1 || echo "Successfully restarted k8s services.";'
```
MarcelMue: Can we check if this `exit 1` statement actually gets hit now? Can we somehow make it simpler to ensure we fail? e.g. by simplifying the expression here?

service/controller/v26/cloudconfig/master_template.go
```diff
@@ -213,8 +213,8 @@ Requires=k8s-kubelet.service
 After=k8s-kubelet.service
 
 [Service]
-Environment="KUBECONFIG=/etc/kubernetes/config/addons-kubeconfig.yml"
-Environment="KUBECTL=quay.io/giantswarm/docker-kubectl:e777d4eaf369d4dabc393c5da42121c2a725ea6a"
+Environment="KUBECONFIG=/etc/kubernetes/kubeconfig/addons.yaml"
+Environment="KUBECTL=quay.io/giantswarm/docker-kubectl:0df981ec7d2bd3c1191f75a0890f1089762febfb"
 ExecStart=/bin/sh -c '\
 	while [ "$(/usr/bin/docker run -e KUBECONFIG=${KUBECONFIG} --net=host --rm -v /etc/kubernetes:/etc/kubernetes $KUBECTL get cs | grep Healthy | wc -l)" -ne "3" ]; do sleep 1 && echo "Waiting for healthy k8s";done;sleep 30s; \
```
MarcelMue: Same with this line. Seems way too complicated for what it is actually trying to do.

MarcelMue: There should be a number of retries here (however high) to prevent us from hitting the same issue again.

tfussell: Yeah, I can use the same retry logic as is used in the inner statement to make sure it fails loudly instead of retrying forever.

service/controller/v26/cloudconfig/master_template.go
tfussell: I think it's not an error using `kubectl delete pod -l k8s-app=coredns` even if no pod matches that label. I will confirm though.

tfussell:
I agree with Joe that the interesting part is how this was not noticed. The first part of the answer is that the first while loop continued forever instead of failing because it didn't have a timeout so we didn't get an alert about a failed unit. The second part is that this may no longer be necessary or is only necessary in rare cases. I did some digging through git history and found https://github.com/projectcalico/calico/issues/1420 which seems to be the original reason for this. Not sure how complicated it is to test.

tfussell:
I agree with Joe that the interesting part is how this was not noticed. The first part of the answer is that the first while loop continued forever instead of failing because it didn't have a timeout so we didn't get an alert about a failed unit. The second part is that this may no longer be necessary or is only necessary in rare cases. I did some digging through git history and found https://github.com/projectcalico/calico/issues/1420 which seems to be the original reason for this. Not sure how complicated it is to test.

tfussell:
Actually it sounds like this was fixed in Calico 2.6.5. I will remove this unit entirely and test tomorrow to be sure.

calvix:
I think this one was trying to solve the different issues, and that was a problem when you delete masters then often the connection was timeouts because the way how the traffic was handled in LB and ingress. 

Not sure if anything was fixed, we definitely needed it even after 2.6.5  otherwise we would remove it :p

The issue was reproducible by doing a master node reboot. (deleting master pod) and then if you create/spawn new pods they had connectivity issues (calico failure) or was not reachable by sv ( kube-proxy failure).

tfussell:
> I think this one was trying to solve the different issues, and that was a problem when you delete masters then often the connection was timeouts because the way how the traffic was handled in LB and ingress.
> 
> Not sure if anything was fixed, we definitely needed it even after 2.6.5 otherwise we would remove it :p
> 
> The issue was reproducible by doing a master node reboot. (deleting master pod) and then if you create/spawn new pods they had connectivity issues (calico failure) or was not reachable by sv ( kube-proxy failure).

Thanks for the context. I will try to reproduce when I have some free time.

tfussell:
I'm updating this systemd unit in https://github.com/giantswarm/kvm-operator/pull/965 so I'm going to close this one. It's still not clear if the bug it fixes is still around, but it doesn't hurt to have it and we will continue to improve when we have the time.

PR #778, opened 2019-11-25
Backporting label changes

tomahawk28:
Toward giantswarm/giantswarm#7763

Applying https://github.com/giantswarm/kvm-operator/pull/776 into all versions.


rossf7 (commented):
I have a concern about backporting to very old versions.

We only started creating app CRs in 0.20.0 of cluster-operator which is in KVM release `9.0.0`.
https://github.com/giantswarm/releases/blob/0f91d51dc09b03a29ba1f0d06d79747550b5a03d/kvm.yaml#L24-L46

So I think we only need to backport to v25. WDYT?

tomahawk28:
So set aside of `app` CR, no one using `*-cluster-values`? That means we could revert older than `v25` changes.

rossf7:
> So set aside of app CR, no one using *-cluster-values? That means we could revert older than v25 changes.

Yes the configmap is only for use by app CRs. So we don't need to change the older versions.

rossf7 (changes requested):
@tomahawk28 Sorry for the hassle about removing changes but the diff is borked. PTAL

tomahawk28:
@rossf7 Ooops, sorry,

rossf7 (approved):
LGTM

PR #779, opened 2019-11-25
Getting current configmap by using labels

tomahawk28:
Toward giantswarm/giantswarm#7763

### DIFF
--- a/service/controller/v26/resource/configmap/current.go
+++ b/service/controller/v26/resource/configmap/current.go
@@ -9,6 +9,8 @@ import (
 	apiv1 "k8s.io/api/core/v1"
 	apismetav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	"github.com/giantswarm/kvm-operator/pkg/label"
+	"github.com/giantswarm/kvm-operator/pkg/project"
 	"github.com/giantswarm/kvm-operator/service/controller/v26/key"
 )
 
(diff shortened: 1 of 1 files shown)


service/controller/v26/resource/configmap/current.go
```diff
@@ -31,7 +33,12 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 	var currentConfigMaps []*apiv1.ConfigMap
 	{
 		namespace := key.ClusterNamespace(customResource)
-		configMapList, err := r.k8sClient.CoreV1().ConfigMaps(namespace).List(apismetav1.ListOptions{})
+
+		listOptions := apismetav1.ListOptions{
```
tomahawk28: After backporting this change then I will replace `apimetav1` into `metav1`.

service/controller/v26/resource/configmap/current.go
```diff
@@ -31,7 +33,12 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
...
+			LabelSelector: fmt.Sprintf("%s=%s", label.ManagedBy, project.Name()),
```
tomahawk28: After this, kvm-operator should not delete configmaps from cluster-operator.

tomahawk28:
Tested on geckon

service/controller/v26/resource/configmap/current.go
rossf7:
```suggestion
		lo := apismetav1.ListOptions{
```

Not a blocker but I think this can be shorter.

rossf7 (approved):
LGTM

PR #783, opened 2019-11-28
Backporting v24 to have proper labels in configmaps. 

tomahawk28:

Many installations still using 8.4.0 release which linked to v24 in kvm-operator. 
It still doesn't have our fix from labeling.

### DIFF
--- a/service/controller/v24/resource/configmap/desired.go
+++ b/service/controller/v24/resource/configmap/desired.go
@@ -9,6 +9,8 @@ import (
 	apiv1 "k8s.io/api/core/v1"
 	apismetav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	"github.com/giantswarm/kvm-operator/pkg/label"
+	"github.com/giantswarm/kvm-operator/pkg/project"
 	"github.com/giantswarm/kvm-operator/service/controller/v24/key"
 )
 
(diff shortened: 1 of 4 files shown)


service/controller/v24/resource/configmap/desired.go
```diff
@@ -96,8 +98,13 @@ func (r *Resource) newConfigMap(customResource v1alpha1.KVMConfig, template stri
 			ObjectMeta: apismetav1.ObjectMeta{
 				Name: key.ConfigMapName(customResource, node, prefix),
 				Labels: map[string]string{
-					"cluster":  key.ClusterID(customResource),
-					"customer": key.ClusterCustomer(customResource),
+					// TODO: Delete two legacy labels from next release
```
MarcelMue: Why is this TODO here when you are backporting?

tomahawk28: it's on v25, v26 already. Do you think it's enough to have TODO only on WIP?

rossf7: Yes lets remove it. We are only backporting because the bug affects production clusters. 

We won't fix v24 so better to just have the comment in WIP versions.

rossf7 (approved):
LGTM once TODO is removed

PR #784, opened 2019-11-28
Backporting get configmaps by labels into v24

tomahawk28:

(Last patch)

Backporting a patch into v24 so only fetch configmaps with proper labels.


rossf7 (approved):
LGTM

PR #792, opened 2019-12-09
Revert IPVS

mazzy89:


tfussell:
This seems fine, but it would be nice to add a short description of why we're making this change. Maybe link to the issue. And don't forget to assign yourself!

mazzy89:
This revert comes out of the discussion we had inside the team where we decided to rollback IPVS and don't release it in the next GS release.

I'll report a more detailed discussion in here https://github.com/giantswarm/giantswarm/issues/6562

PR #801, opened 2019-12-13
Using project package

tomahawk28:
Align with other operators for `project` package usage.

### DIFF
--- a/service/service.go
+++ b/service/service.go
@@ -18,19 +18,18 @@ import (
 	"k8s.io/client-go/rest"
 
 	"github.com/giantswarm/kvm-operator/flag"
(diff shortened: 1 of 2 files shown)


service/service.go
```diff
@@ -160,7 +166,7 @@ func New(config Config) (*Service, error) {
 			TenantCluster: tenantCluster,
 
 			CRDLabelSelector: config.Viper.GetString(config.Flag.Service.CRD.LabelSelector),
-			ProjectName:      config.Name,
+			ProjectName:      project.Name(),
```
tomahawk28: Rather than having each project package on each controller file, I think this is cleaner.

rossf7 (approved):
Nice! LGTM

PR #802, opened 2019-12-16
Filter by labels using new operatorkit

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/8129

By using `MatchLabel`, controller would only work with objects with matching labels.


rossf7 (approved):
LGTM 🚀 

Thanks for getting this sorted.

PR #803, opened 2019-12-17
Better error handling on k8sclient creation

tomahawk28:
Unfortunately, I deleted the logs from kvm-operator, but it put an error stack logs on k8sclient if we just created TCs. 

This PR would cancel the `node` resource gracefully.


tomahawk28:
Tested on `geckon`

tomahawk28:
Do you guys think we need to backporting all this?

sslavic:
I think it has to be applied to all supported versions in https://github.com/giantswarm/kvm-operator/tree/master/service/controller just like k8s dependency updates and all refactoring due to internal library updates was applied.

xh3b4sd (approved):

PR #810, opened 2020-01-31
Test go modules

tfussell:
This will let us keep track of internal libraries that haven't been migrated to go modules yet. Once all giantswarm deps have a published release pinned in this `go.mod`, it should be ready to merge.


stone-z:
Done in https://github.com/giantswarm/kvm-operator/pull/847

PR #822, opened 2020-02-05
Use operator version

MarcelMue:
Main difference here is that it is now using the label instead of the VBV spec field.

I was also so free and renamed the function to match `aws-operator` master in the same go.

towards https://github.com/giantswarm/giantswarm/issues/7820


MarcelMue:
Tested & works.

PR #824, opened 2020-02-06
flattening: various refactoring for master

tfussell:
Just importing `project.Name()` where it's needing instead of injecting it throughout the program.


tfussell:
This is a relic from flattening. May as well get it in though.

PR #839, opened 2020-03-20
Prepare helm for flat deployment

stone-z:


tfussell:
Can we go ahead and rename `helm/kvm-operator-chart` to `helm/kvm-operator` here instead of creating a second parallel chart? Would be a little nicer to viewing differences, but I don't totally remember the process we used before.

stone-z:
> Can we go ahead and rename `helm/kvm-operator-chart` to `helm/kvm-operator` here instead of creating a second parallel chart? Would be a little nicer to viewing differences, but I don't totally remember the process we used before.

Sure, I thought we needed both for e2e tests. Will update momentarily

stone-z:
Will rename the folder back to `helm/kvm-operator` once the changes are approved

tfussell:
> > Can we go ahead and rename `helm/kvm-operator-chart` to `helm/kvm-operator` here instead of creating a second parallel chart? Would be a little nicer to viewing differences, but I don't totally remember the process we used before.
> 
> Sure, I thought we needed both for e2e tests. Will update momentarily

I think you're right except we don't have e2e tests on this operator as far as I know.

PR #840, opened 2020-03-20
Update CircleCI to match master

stone-z:
Same as master -- anything special needed for legacy branch?


tfussell:
Looks fine. Just need to rename `helm/kvm-operator-chart` to `helm/kvm-operator` I think.

tfussell:
I guess this is fixed in https://github.com/giantswarm/kvm-operator/pull/839 so this should be fine if you merge that and then update this branch. Up to you.

PR #843, opened 2020-03-31
Get from releases

stone-z:


Gopkg.toml
```diff
@@ -100,37 +100,37 @@ required = [
   name = "github.com/spf13/viper"
   version = "1.0.0"
 
-[[constraint]]
+[[override]]
   name = "k8s.io/api"
-  version = "kubernetes-1.16.3"
+  version = "kubernetes-1.15.11"
```
MarcelMue: Shouldn't be necessary these are backwards compatible.

pkg/project/version_bundle.go
```diff
@@ -8,23 +8,8 @@ func NewVersionBundle() versionbundle.Bundle {
...
-			{
-				Component:   "etcd",
-				Description: "Updated from v3.3.13 to v3.3.15.",
-				Kind:        versionbundle.KindChanged,
-			},
-			{
-				Component:   "kubernetes",
-				Description: "Updated from v1.14.6 to v1.15.5.",
+				Component:   "kvm-operator",
+				Description: "Modified to retrieve component versions from releases",
```
MarcelMue: This is unused - fine as a placeholder but we need a `CHANGELOG.md`

service/controller/cloudconfig/master_template.go
```diff
@@ -7,29 +7,28 @@ import (
...
 	"github.com/giantswarm/microerror"
-	"github.com/giantswarm/randomkeys"
 
 	"github.com/giantswarm/kvm-operator/service/controller/key"
 )
 
 // NewMasterTemplate generates a new worker cloud config template and returns it
 // as a base64 encoded string.
-func (c *CloudConfig) NewMasterTemplate(customObject v1alpha1.KVMConfig, certs certs.Cluster, node v1alpha1.ClusterNode, randomKeys randomkeys.Cluster, nodeIndex int) (string, error) {
+func (c *CloudConfig) NewMasterTemplate(data IgnitionTemplateData, node v1alpha1.ClusterNode, nodeIndex int) (string, error) {
```
MarcelMue: This is one part of it here but we also need to alter the places where these functions are called - or is this missing in the diff somehow?

pkg/project/version_bundle.go
stone-z: Added Changelog

Gopkg.toml
stone-z: Updated to newer 1.16.8 due to vulnerable dependencies in 1.16.3

service/controller/cloudconfig/master_template.go
stone-z: Yep, just hadn't gotten there yet. Has been added

service/controller/resource/configmap/desired.go
```diff
@@ -51,7 +52,14 @@ func (r *Resource) newConfigMaps(customResource v1alpha1.KVMConfig) ([]*apiv1.Co
 			return nil, microerror.Maskf(notFoundError, fmt.Sprintf("node index for master (%q) is not available", node.ID))
 		}
 
-		template, err := r.cloudConfig.NewMasterTemplate(customResource, certs, node, keys, nodeIdx)
+		ignition := cloudconfig.IgnitionTemplateData{
+			CustomObject: customResource,
+			ClusterCerts: certs,
+			ClusterKeys:  keys,
+			// Images?
```
MarcelMue, line 59: Images!

You can check `aws-operator` on how to actually get the images here. We can also hang tomorrow if you want.

stone-z: I checked there but I think this works differently for KVM?

MarcelMue: It should be reasonably similar :)

tfussell:
Looking good so far. Just a failing unit test now it seems.

.dockerignore
```diff
@@ -0,0 +1,3 @@
+*
+!kvm-operator
+!go.mod
```
MarcelMue, line 3: Why is a dockerignore needed?

service/controller/resource/configmap/desired.go
```diff
@@ -45,13 +48,37 @@ func (r *Resource) newConfigMaps(customResource v1alpha1.KVMConfig) ([]*apiv1.Co
 		return nil, microerror.Mask(err)
 	}
 
+	var release *releasev1alpha1.Release
+	{
+		releaseVersion := customResource.Labels[label.ReleaseVersion]
+		releaseName := fmt.Sprintf("v%s", releaseVersion)
+		release, err = r.g8sClient.ReleaseV1alpha1().Releases().Get(releaseName, apismetav1.GetOptions{})
```
MarcelMue, line 55: RBAC needs to be changed to allow us getting `Releases`

.dockerignore
stone-z: I'm actually not convinced that it is anymore -- removing to see what happens

service/controller/resource/configmap/desired.go
stone-z: Added

service/controller/resource/configmap/desired.go
stone-z: It was :)

stone-z:
I think this is good to go. I'll do flatcar in a separate PR

go.mod
```diff
@@ -0,0 +1,39 @@
...
+	github.com/giantswarm/errors v0.2.0
+	github.com/giantswarm/k8sclient v0.2.0
+	github.com/giantswarm/k8scloudconfig/v6 v6.0.1
+	github.com/giantswarm/microendpoint v0.2.0
+	github.com/giantswarm/microerror v0.2.0
+	github.com/giantswarm/microkit v0.2.0
+	github.com/giantswarm/micrologger v0.3.1
+	github.com/giantswarm/operatorkit v0.2.0
+	github.com/giantswarm/randomkeys v0.2.0
+	github.com/giantswarm/statusresource v0.0.0-20200103225110-1c017ce979b3
```
MarcelMue, line 17: We can improve here later on probably this `go.mod` doesn't look super clean :D

service/controller/resource/configmap/desired.go
```diff
@@ -45,13 +48,37 @@ func (r *Resource) newConfigMaps(customResource v1alpha1.KVMConfig) ([]*apiv1.Co
 		return nil, microerror.Mask(err)
 	}
 
+	var release *releasev1alpha1.Release
+	{
+		releaseVersion := customResource.Labels[label.ReleaseVersion]
```
MarcelMue, line 53: Nit: I would like this line and the next to be in some `key.go` but can be done in a follow up :)

PR #846, opened 2020-04-02
Pull Flatcar version from releases

stone-z:


.dependabot/config.yml
```diff
@@ -0,0 +1,11 @@
+version: 1
```
tfussell, line 1: Does this work correctly on a branch or does this need to be on master?

stone-z: You're right, I think this has to be on master. Will add it there

go.mod
```diff
@@ -5,7 +5,7 @@ go 1.14
 require (
 	github.com/giantswarm/apiextensions v0.2.0
 	github.com/giantswarm/certs v0.2.0
-	github.com/giantswarm/errors v0.2.1
+	github.com/giantswarm/errors v0.2.2-0.20200406104130-71836ca40924
```
MarcelMue, line 8: I think you can use latest now.

MarcelMue: NVM - just saw your PR.

.dependabot/config.yml
```diff
@@ -0,0 +1,11 @@
...
+update_configs:
+  - package_manager: "go:modules"
+    directory: "/"
+    target_branch: "legacy-1-15"
+    update_schedule: "daily"
```
MarcelMue, line 6: I think even less frequent would be fine.

go.mod
```diff
@@ -5,16 +5,16 @@ go 1.14
 require (
 	github.com/giantswarm/apiextensions v0.2.0
 	github.com/giantswarm/certs v0.2.0
-	github.com/giantswarm/errors v0.2.0
+	github.com/giantswarm/errors v0.2.2
 	github.com/giantswarm/k8sclient v0.2.0
-	github.com/giantswarm/k8scloudconfig/v6 v6.0.1
+	github.com/giantswarm/k8scloudconfig/v6 v6.0.3-0.20200414150239-6351b9c5f05e
```
MarcelMue, line 10: Please make sure to revert this to a tag at the end.

service/controller/resource/configmap/desired.go
```diff
@@ -63,6 +65,11 @@ func (r *Resource) newConfigMaps(customResource v1alpha1.KVMConfig) ([]*apiv1.Co
 		return nil, microerror.Mask(err)
 	}
 
+	defaultVersions := key.DefaultVersions()
```
MarcelMue, line 68: My 2c:
I don't like these defaultVersion things - Why not just create the struct and set all values?

stone-z: I thought this was a placeholder until ignition-operator took over the rendering. I'm not a huge fan of the implementation here but I do like the idea of keeping some sane defaults in one place as opposed to scattered across the different resources

.dependabot/config.yml
stone-z: This will be removed from this PR as it is already on master anyway. But we should at least get checked for security updates daily. That is also configurable on master

PR #847, opened 2020-04-08
Migrate to go modules

stone-z:


stone-z:
This is running on geckon

Dockerfile
```diff
@@ -1,11 +1,18 @@
+FROM golang:1.13 AS builder
```
tfussell, line 1: It would be nice to use our retagged golang image and 1.14.1 to match `architect`.

stone-z: Good thinking - updated

PR #857, opened 2020-04-09
CI updates

stone-z:
- Add dependabot config
- Update architect orb
- fix linting issues - remove unused code, catch error return, simplify

### DIFF
--- a/helm/kvm-operator/ci/default-values.yaml
+++ b/helm/kvm-operator/ci/default-values.yaml
@@ -3,5 +3,5 @@ Installation:
   V1:
     Secret:
       Registry:
-      PullSecret:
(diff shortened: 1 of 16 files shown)


stone-z:
Will update the required checks for this repo once merged

stone-z:
Running on geckon

helm/kvm-operator/ci/default-values.yaml
```diff
@@ -0,0 +1,31 @@
+---
```
MarcelMue, line 1: Why is this needed?

stone-z: The file is needed to keep the new helm chart linting happy

The line might not be necessary, I just based this off of what's in use elsewhere

PR #865, opened 2020-04-09
Get flatcar version from releases

stone-z:
Deployed on gorgoth, create, scale, delete operations all working

### DIFF
--- a/service/controller/cluster.go
+++ b/service/controller/cluster.go
@@ -104,6 +104,7 @@ func newClusterResourceSets(config ClusterConfig) ([]*controller.ResourceSet, er
 	{
 		c := ClusterResourceSetConfig{
 			CertsSearcher:      config.CertsSearcher,
+			G8sClient:          config.K8sClient.G8sClient(),
 			K8sClient:          config.K8sClient,
 			Logger:             config.Logger,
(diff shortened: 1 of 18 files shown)


service/controller/cluster.go
```diff
@@ -104,6 +104,7 @@ func newClusterResourceSets(config ClusterConfig) ([]*controller.ResourceSet, er
 	{
 		c := ClusterResourceSetConfig{
 			CertsSearcher:      config.CertsSearcher,
+			G8sClient:          config.K8sClient.G8sClient(),
```
tfussell, line 107: Since you're passing the full K8sClient into this config as well, you could probably drop this line and get the G8sClient from it inside.

MarcelMue:
Fine with me if tested :)

service/controller/cluster.go
stone-z: I kind of like exposing both clients in this case because it isn't necessarily obvious that a K8sClient would contain a G8sClient, and this way it matches the format of all the other constructors. Going to go with this but open to discussion if someone wants

tfussell: I think it's good to be consistent with passing in only the child `G8sClient`/`K8sClient` or the parent `K8sClient`, but no worries if you prefer this way.

PR #870, opened 2020-04-14
Get versions from releases

stone-z:


stone-z:
Worker comes up, master does not -- missing environment files, e.g.
`master-r4vr6-64565f5b57-kxq4q login: [  368.850492] systemd[1]: etcd3.service: Failed to load environment files: No such file or directory`

service/controller/key/key.go
```diff
@@ -104,6 +105,13 @@ const (
 	LegacyLabelCluster = "cluster"
 )
 
+const (
+	// TODO: Change to 1.16.8
+	kubectlVersion               = "9ccdc9dc55a01b1fde2aea73901d0a699909c9cd" // 1.15.5
+	KubernetesNetworkSetupDocker = "9ccdc9dc55a01b1fde2aea73901d0a699909c9cd"
```
tfussell, line 111:
```suggestion
	KubernetesNetworkSetupDocker = "68e90113331feca3b9ffe6a75a601b381ba8c1f7"
```

stone-z: Was already on the way :)

tfussell: Nice--didn't meant to step on your toes. It just caught my eye that the tags were identical between this and `kubectl` above.

stone-z: No worries, thanks for the help. Trying with the new version now

tfussell: All three images are at `68e90113331feca3b9ffe6a75a601b381ba8c1f7` now. That's probably not intended.

stone-z: Cluster is up :) those were pinned to the same version originally so I aligned them again

tfussell: Sounds good.

stone-z:
Cluster creates, scales, and deletes. Update issue will be handled in separate PR

stone-z:
@giantswarm/team-ludacris going with this if there are no objections

PR #871, opened 2020-04-15
Add supports to set the OIDC username and groups prefix

omegas27:
Towards https://github.com/giantswarm/giantswarm/issues/10012

@giantswarm/team-ludacris I set that as a draft because I would need someones help to test it.

I will document it somewhere afterwards

### DIFF
--- a/service/controller/cluster.go
+++ b/service/controller/cluster.go
@@ -40,7 +40,7 @@ type ClusterConfigOIDC struct {
 	UsernameClaim  string
 	UsernamePrefix string
 	GroupsClaim    string
-	GroupsPrefix    string
+	GroupsPrefix   string
 }
 
 type Cluster struct {
--- a/service/controller/cluster.go
+++ b/service/controller/cluster.go
@@ -35,10 +35,12 @@ type ClusterConfig struct {
 // ClusterConfigOIDC represents the configuration of the OIDC authorization
 // provider.
 type ClusterConfigOIDC struct {
-	ClientID      string
-	IssuerURL     string
-	UsernameClaim string
-	GroupsClaim   string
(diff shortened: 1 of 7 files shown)


service/controller/cluster.go
```diff
@@ -118,10 +118,12 @@ func newClusterResourceSets(config ClusterConfig) ([]*controller.ResourceSet, er
 			NTPServers:         config.NTPServers,
 			ProjectName:        config.ProjectName,
 			OIDC: cloudconfig.OIDCConfig{
```
MarcelMue, line 120: You will need to re-generate the golden files in these unit tests to make this build.

omegas27: Sure :) 
How can I do that?

MarcelMue: Actually that doesn't seem to be the issue - let me check. It builds for me locally

omegas27: :+1:

MarcelMue: @giantswarm/team-ludacris Can someone take a look here? I can not figure out why this build is failing.

tfussell: Maybe the CircleCI runner is OOM? https://forum.golangbridge.org/t/go-build-exits-with-signal-killed/513

stone-z: I had this repeatedly with kvm-operator too, intermittently. Retrying the build worked, not sure what the root cause is unless it is OOM or something Circle-side (my failures were right after a documented Circle outage so I guessed it was related)

omegas27: It actually passed after a re-run

PR #892, opened 2020-04-17
Recreate crbs

stone-z:
Adding the CRBs rolls the nodes as-is. This leaves the old CRB installed, I will address how and when to delete them in a separate PR


stone-z:
Wow that commit history is hideous. Will do this again

PR #893, opened 2020-04-17
Catch error when RoleRef can't be changed and create a new CRB instead

stone-z:
This leaves the old CRB in-place. I will revisit if/how/when to delete old CRBs in a followup PR


service/controller/resource/clusterrolebinding/update.go
```diff
@@ -21,7 +22,19 @@ func (r *Resource) ApplyUpdateChange(ctx context.Context, obj, updateChange inte
 		// Create the cluster role bindings in the Kubernetes API.
 		for _, clusterRoleBinding := range clusterRoleBindingsToUpdate {
 			_, err := r.k8sClient.RbacV1().ClusterRoleBindings().Update(clusterRoleBinding)
-			if err != nil {
+			if isExternalFieldImmutableError(err) {
+				// We can't change a RoleRef, so create a new CRB which references the new RoleRef
+				newCRB := apiv1.ClusterRoleBinding{}
+
+				clusterRoleBinding.DeepCopyInto(&newCRB)
+				newCRB.SetName(clusterRoleBinding.Name + "-upgrading")
```
MarcelMue, line 30: We should change this back eventually somehow. Possibly automatically?

stone-z: Yeah agreed, I'm not sure yet exactly when it is "safe" to delete the old one. I can wait until all old pods are gone, but I don't know how/where to track this at the moment

stone-z: Actually, I don't think I can rename the existing CRB. So the process would be 1 create the new one with '-upgrading', 2 delete the old one, 3 create a new new one without the suffix, 4 delete the '-upgrading' CRB. I think that would result in a second roll of the nodes

stone-z: Otherwise I YOLO it and delete the old one before creating the new one

tfussell: After thinking about this a bit, the important thing is that we can't change the `ClusterRole` of a `ClusterRoleBinding` from `kvm-operator` to `kvm-operator-<version>` or from `kvm-operator-<previous version>` to `kvm-operator-<version>`. In short, this will happen any time the operator version is changed.

Instead of appending `-upgrading` to the name of the CRB and hitting this same issue on the next upgrade, what do you think of instead appending the `ClusterRole` name to it or a short hash thereof (maybe using crc32)? For TC with ID `abc12`, the CRB would look like `abc12-kvm-operator-<version>` or `abc12-<hash>`. Once the rollout of the new nodes is complete, we can then look for and delete any CRB matching `abc12*` other than the CRB currently in use. This is similar to how Kubernetes manages a `StatefulSet` based on names. We could alternatively select CRBs by label `giantswarm.io/cluster-id=abc12` to clean them up.

MarcelMue: I agree that creating one from e.g. the operator version would make a lot of sense.

stone-z: Yes also agreed. However it may not even be necessary. I just tested an update using a version which creates a temporary CRB, deletes the original, then creates the new CRB using the old name. So the end effect looks like it was updated in-place. The upgrade was successful and all TC pods are running, so initial impression is ok. Will update here momentarily, then we can decide if there is still a need to add versioned CRBs

service/controller/resource/clusterrolebinding/update.go
```diff
@@ -21,7 +23,43 @@ func (r *Resource) ApplyUpdateChange(ctx context.Context, obj, updateChange inte
...
+				// Try deleting and then move to re-create-crb branch
+				// Delete the old CRB
+				err = r.k8sClient.RbacV1().ClusterRoleBindings().Delete(clusterRoleBinding.Name, &apismetav1.DeleteOptions{})
+				if apierrors.IsNotFound(err) {
+				} else if err != nil {
+					return microerror.Mask(err)
+				}
+
+				// Create a CRB with the original name, but use the new configuration
+				newCRB.SetName(clusterRoleBinding.Name)
```
MarcelMue, line 48: We talked and this name should be the one coming from the `helm` template.

It seems like from testing that the `-upgrading` one is not needed at all. If I understood @stone-z correctly.

stone-z:
I commented out the ClusterRoleBindings from the resource set and retried the upgrade. The cluster upgraded and all pods are running. I don't think these are actually needed anymore - the operator's own binding is handled by helm

MarcelMue:
> I commented out the ClusterRoleBindings from the resource set and retried the upgrade. The cluster upgraded and all pods are running. I don't think these are actually needed anymore - the operator's own binding is handled by helm

Please keep some of the logic to clean up the old bindings - that way we clean them up at least.

Additionally it would be good to ask in slack - maybe #sig-infra or #sig-operator to see if we are missing something obvious here.

tfussell:
> I commented out the ClusterRoleBindings from the resource set and retried the upgrade. The cluster upgraded and all pods are running. I don't think these are actually needed anymore - the operator's own binding is handled by helm

I think the CRB is required for the PSP to be applied to the TC node pods at least. I can imagine that the pods also need some of the RBAC permissions though I'm not sure why they would share the `ClusterRole` of the parent `kvm-operator`.

stone-z:
@tfussell I thought the same thing but I couldn't find anything in the TCs.

HOWEVER I removed the CRB and now k8s-endpoint-updater complains, so I think we do actually still need it for that

service/controller/resource/clusterrolebinding/update.go
```diff
@@ -21,7 +23,24 @@ func (r *Resource) ApplyUpdateChange(ctx context.Context, obj, updateChange inte
 		// Create the cluster role bindings in the Kubernetes API.
 		for _, clusterRoleBinding := range clusterRoleBindingsToUpdate {
 			_, err := r.k8sClient.RbacV1().ClusterRoleBindings().Update(clusterRoleBinding)
-			if err != nil {
+			if isExternalFieldImmutableError(err) {
+				// We can't change a RoleRef, so delete the old CRB and replace it
```
MarcelMue, line 27: Maybe we should add a `log` message here as well.

stone-z: Good idea - added some

stone-z:
> LGTM when build goes green.

I think this is related to architect changes. Moving this to the orb in https://github.com/giantswarm/kvm-operator/pull/902

MarcelMue:
> > LGTM when build goes green.
> 
> I think this is related to architect changes. Moving this to the orb in #902

Should be fixed - try again :)

PR #903, opened 2020-04-23
Release v3.9.1

stone-z:
Update Changelog and set operator version. Component versions are now pulled from releases


tfussell:
It looks like `v25` (which this is based on?) was using k8scc `v_4_9_0` and this release is using `v_4_9_2`, so maybe we should mention these two changes in the operator from `v_4_9_1` https://github.com/giantswarm/k8scloudconfig/blob/master/CHANGELOG.md#v491---2020-03-10

stone-z:
> maybe we should mention these two changes in the operator from `v_4_9_1` 

@tfussell yes we should -- added

PR #922, opened 2020-04-27
Non empty changelogs

MarcelMue:
Apparently changelogs can't be empty. Am I missing a vendor @tfussell ?

### DIFF
--- a/pkg/project/version_bundle.go
+++ b/pkg/project/version_bundle.go
@@ -6,6 +6,16 @@ import (
 
 func NewVersionBundle() versionbundle.Bundle {
 	return versionbundle.Bundle{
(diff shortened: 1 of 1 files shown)


tfussell:
> Apparently changelogs can't be empty. Am I missing a vendor @tfussell ?

No. We need to drop that requirement from `versionbundle` or even drop the dependency altogether. For now we've just kept a single placeholder change in each operator.

MarcelMue:
> > Apparently changelogs can't be empty. Am I missing a vendor @tfussell ?
> 
> No. We need to drop that requirement from `versionbundle` or even drop the dependency altogether. For now we've just kept a single placeholder change in each operator.

Ack - I might provide a PR to allow us to drop them.

PR #923, opened 2020-04-27
Remove CRD management

MarcelMue:
Do we also need to add something to opsctl or is this fine for now?

### DIFF
--- a/service/controller/cluster.go
+++ b/service/controller/cluster.go
@@ -63,7 +63,6 @@ func NewCluster(config ClusterConfig) (*Cluster, error) {
 	var operatorkitController *controller.Controller
 	{
 		c := controller.Config{
-			CRD:          v1alpha1.NewKVMConfigCRD(),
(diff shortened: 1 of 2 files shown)


MarcelMue:
@giantswarm/team-ludacris PTAL as I want to tag kvm-operator :D

stone-z:
Should this also be removed elsewhere? e.g. flag, the helm configmap, and service.go?

MarcelMue:
> Should this also be removed elsewhere? e.g. flag, the helm configmap, and service.go?

I don't think so - this should match the change @tfussell did on AWS operator.

tfussell:
> Do we also need to add something to opsctl or is this fine for now?

Sorry for being slow looking at this. We'll depend on released operators to ensure the CRDs for now. Once we have migrated KVM CRDs to v1, we will uncomment it in `opsctl ensure crds`, then deploy our patched operators.

PR #928, opened 2020-04-28
Use Release.Revision in chart for Helm 3 support

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/10098

I missed kvm-operator but we need this for Helm 3 and to allow testing of the control plane upgrade.


rossf7:
Tested on `gorgoth`.

PR #929, opened 2020-04-28
Use k8scc with binaries from hyperkube

MarcelMue:


tfussell:
I think we can close this one.

PR #934, opened 2020-04-28
Add github action for dependabot on legacy-1-15

stone-z:
Same as on master. This must exist to trigger the action on branches from legacy-1-15


stone-z:
Going with a ping

PR #970, opened 2020-05-20
Fix cluster creation using float64 wrapper

tfussell:


service/controller/key/key.go
```diff
@@ -242,8 +242,8 @@ func DockerVolumeSizeFromNode(node v1alpha1.KVMConfigSpecKVMNode) string {
 		return fmt.Sprintf("%dG", node.DockerVolumeSizeGB)
 	}
 
-	if node.Disk != "" {
-		return fmt.Sprintf("%sG", node.Disk)
+	if node.Disk.Value != 0 {
+		return fmt.Sprintf("%sG", strconv.FormatFloat(node.Disk.Value, 'f', 0, 64))
```
tfussell, line 246: Reverting to older format function from https://github.com/giantswarm/kvm-operator/blob/3c73f5cbfc2aa73ca902cfbe1ee9f7811f63bd90/service/controller/key/key.go#L218

MarcelMue:
Merging here so I can test this today.

PR #976, opened 2020-05-26
Check tc master status before upgrading workers

stone-z:
Towards https://github.com/giantswarm/giantswarm/issues/9760

### DIFF
--- a/pkg/project/project.go
+++ b/pkg/project/project.go
@@ -5,7 +5,7 @@ var (
 	gitSHA             = "n/a"
 	name        string = "kvm-operator"
 	source      string = "https://github.com/giantswarm/kvm-operator"
-	version            = "3.11.3-dev"
+	version            = "3.11.2-dev"
 )
 
 func Description() string {
--- a/pkg/project/project.go
+++ b/pkg/project/project.go
@@ -5,7 +5,7 @@ var (
 	gitSHA             = "n/a"
 	name        string = "kvm-operator"
 	source      string = "https://github.com/giantswarm/kvm-operator"
-	version            = "3.11.2-dev"
+	version            = "3.11.3-dev"
 )
 
 func Description() string {
--- a/service/controller/resource/deployment/update.go
+++ b/service/controller/resource/deployment/update.go
@@ -4,10 +4,15 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/giantswarm/errors/tenant"
+	"github.com/giantswarm/k8sclient"
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/operatorkit/controller/context/updateallowedcontext"
 	"github.com/giantswarm/operatorkit/resource/crud"
+	"github.com/giantswarm/tenantcluster"
 	v1 "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
 
 	"github.com/giantswarm/kvm-operator/service/controller/key"
 )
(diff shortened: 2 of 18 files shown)


stone-z:
The ifs for nil tenantclusters will be removed prior to merging -- this was to allow me to test locally without faking a client

service/controller/resource/deployment/update.go
```diff
@@ -73,6 +79,47 @@ func (r *Resource) newUpdateChange(ctx context.Context, obj, currentState, desir
 		return nil, microerror.Mask(err)
 	}
 
+	// Create a client for the reconciled tenant cluster
+	var tcK8sClient kubernetes.Interface
+	{
+		if r.tenantCluster != nil {
```
stone-z, line 85: Will be removed

service/controller/resource/deployment/update.go
```diff
@@ -110,6 +157,32 @@ func (r *Resource) newUpdateChange(ctx context.Context, obj, currentState, desir
 				continue
 			}
 
+			// If worker deployment, check that master does not have any prohibited states before updating the worker
+			if tcK8sClient != nil {
```
stone-z, line 161: Will be kept, as the logic depends on it and it may not get created above if the TC does not exist yet

service/controller/resource/deployment/update.go
```diff
@@ -73,6 +79,47 @@ func (r *Resource) newUpdateChange(ctx context.Context, obj, currentState, desir
...
+			}
+
+			tcK8sClient = k8sClients.K8sClient()
+			r.logger.LogCtx(ctx, "level", "debug", "message", "created Kubernetes client for tenant cluster")
+		} else {
+			r.logger.LogCtx(ctx, "level", "warning", "message", "unable to create Kubernetes client for tenant cluster")
+			tcK8sClient = nil
+		}
+	}
+
```
MarcelMue, line 122: Easiest fix for your issue:
Move code from line `124` until `188` into a new private method which has a signature like
```
func (r *Resource) someFunc(ctx context.Context, obj  interface{}, currentDeployments []v1.Deployments, desiredDeployments []v1.Deployments, tcK8sClient kubernetes.Interface ) ([]*v1.Deployment, error)
```
And point the existing unit tests at `someFunc`

service/controller/resource/deployment/update.go
```diff
@@ -73,49 +130,69 @@ func (r *Resource) newUpdateChange(ctx context.Context, obj, currentState, desir
...
+	// We select one deployment to be updated per reconciliation loop. Therefore
+	// we have to check its state on the version bundle level to see if a
+	// deployment is already up to date. We also check if there are any other
+	// changes on the pod specs. In case there are none, we check the next one.
+	// The first one not being up to date will be chosen to be updated next and
+	// the loop will be broken immediately.
+DeploymentsLoop:
```
MarcelMue, line 153: Missing comment?

stone-z: Nope, it's a label

service/controller/resource/deployment/update_test.go
```diff
@@ -44,6 +45,7 @@ func Test_Resource_Deployment_newUpdateChange(t *testing.T) {
 		CurrentState                interface{}
 		DesiredState                interface{}
 		ExpectedDeploymentsToUpdate []*v1.Deployment
+		FakeTCObjects               []runtime.Object
```
MarcelMue, line 48: Need to also change the test name to mirror the change of which function is tested now.

service/controller/resource/deployment/update.go
MarcelMue: Oh - I literally have never used labels in golang before in my life :D

stone-z: Me neither before right now!

service/controller/resource/deployment/update.go
```diff
@@ -73,49 +130,69 @@ func (r *Resource) newUpdateChange(ctx context.Context, obj, currentState, desir
 		return nil, microerror.Mask(err)
 	}
 
-	if updateallowedcontext.IsUpdateAllowed(ctx) {
```
tfussell, line 76: Is it okay to drop this? Not sure what it does.

tfussell: Nevermind, I see it is in the calling function.

service/controller/resource/deployment/update.go
```diff
@@ -73,49 +130,69 @@ func (r *Resource) newUpdateChange(ctx context.Context, obj, currentState, desir
...
+						// Node has NoSchedule or NoExecute taint
+						msg := fmt.Sprintf("not updating deployment '%s': one or more tenant cluster master nodes are unschedulable", currentDeployment.GetName())
+						r.logger.LogCtx(ctx, "level", "warning", "message", msg)
+						continue DeploymentsLoop
+					}
+				}
 			}
+		} else {
+			r.logger.LogCtx(ctx, "level", "warning", "message", "unable to check tenant cluster master status. No tenant cluster client configured")
+			continue DeploymentsLoop
```
tfussell, line 167: I think this could be a plain `continue` since it's not in a nested `for`.

service/controller/resource/deployment/update.go
```diff
@@ -73,49 +130,69 @@ func (r *Resource) newUpdateChange(ctx context.Context, obj, currentState, desir
...
+		if !isDeploymentModified(desiredDeployment, currentDeployment) {
+			r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("not updating deployment '%s': no changes found", currentDeployment.GetName()))
+			continue
+		}
 
-			if !isDeploymentModified(desiredDeployment, currentDeployment) {
-				r.logger.LogCtx(ctx, "level", "debug", "message", fmt.Sprintf("not updating deployment '%s': no changes found", currentDeployment.GetName()))
-				continue
+		// If worker deployment, check that master does not have any prohibited states before updating the worker
+		if tcK8sClient != nil {
```
tfussell, line 148: Can this be `nil`? It seems like reconciliation should stop if the TC is unreachable.

service/controller/resource/deployment/update.go
stone-z: It could be, but I wanted it to be the same as the other continue to be explicit that they both continue to the outer loop. Open to changing it if you have a strong feeling about it

service/controller/resource/deployment/update.go
stone-z: You're right, that was originally for testing. Now the deployment resource enforces the tenant cluster requirement

service/controller/resource/deployment/update.go
stone-z: Actually I removed this altogether as the nil check isn't needed anymore

stone-z:
I have tested this behavior on `geckon` by doing an upgrade and cordoning the master as soon as it comes up. The worker deployment is not updated and the warning is printed

pkg/project/project.go
```diff
@@ -5,7 +5,7 @@ var (
 	gitSHA             = "n/a"
 	name        string = "kvm-operator"
 	source      string = "https://github.com/giantswarm/kvm-operator"
-	version            = "3.11.2-dev"
+	version            = "3.11.3-dev"
```
stone-z, line 8: This is to avoid conflict with other operator test version. Will be reverted

PR #981, opened 2020-06-05
Remove endpoints for non-Ready Pods

stone-z:
Towards https://github.com/giantswarm/giantswarm/issues/11303

### DIFF
--- a/pkg/project/project.go
+++ b/pkg/project/project.go
@@ -5,7 +5,7 @@ var (
 	gitSHA             = "n/a"
 	name        string = "kvm-operator"
 	source      string = "https://github.com/giantswarm/kvm-operator"
-	version            = "3.11.1"
+	version            = "3.11.2-dev"
 )
 
 func Description() string {
--- a/pkg/project/project.go
+++ b/pkg/project/project.go
@@ -5,7 +5,7 @@ var (
 	gitSHA             = "n/a"
 	name        string = "kvm-operator"
 	source      string = "https://github.com/giantswarm/kvm-operator"
-	version            = "3.11.2-dev"
+	version            = "3.11.1"
 )
 
 func Description() string {
--- a/service/controller/resource/cleanupendpointips/create.go
+++ b/service/controller/resource/cleanupendpointips/create.go
@@ -152,20 +152,24 @@ func removeFromEndpointAddressList(addresses []corev1.EndpointAddress, indexesTo
 // removeDeadIPFromEndpoints compares endpoint IPs with current state of nodes and
 // removes any IP addresses that does not belong to any node.
 func removeDeadIPFromEndpoints(endpoints *corev1.Endpoints, nodes []corev1.Node) (int, *corev1.Endpoints) {
+	fmt.Println(fmt.Sprintf("removing dead endpoints from %s", endpoints.Name))
 	endpointAddresses := endpoints.Subsets[0].Addresses
 
 	var indexesToDelete []int
 	for i, ip := range endpointAddresses {
+		fmt.Println(fmt.Sprintf("i: %d // ip: %v#", i, ip))
 		found := false
 		// check if the ip belongs to any k8s node
 		for _, node := range nodes {
-			if node.Labels["ip"] == ip.IP {
+			if node.Labels["ip"] == ip.IP { // Check && node.Status.Conditions
+				fmt.Println(fmt.Sprintf("found IP match"))
 				found = true
 				break
 			}
 		}
 		// endpoint ip does not belong to any node, lets remove it
 		if !found {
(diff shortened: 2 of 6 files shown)


pkg/project/project.go
```diff
@@ -5,7 +5,7 @@ var (
 	gitSHA             = "n/a"
 	name        string = "kvm-operator"
 	source      string = "https://github.com/giantswarm/kvm-operator"
-	version            = "3.11.2-dev"
+	version            = "3.11.1"
```
stone-z, line 8: Will be reverted

service/controller/resource/cleanupendpointips/create.go
```diff
@@ -149,28 +154,58 @@ func removeFromEndpointAddressList(addresses []corev1.EndpointAddress, indexesTo
...
 }
 
+func controlPlanePodForTCNode(node corev1.Node, pods []corev1.Pod) (corev1.Pod, error) {
+	for _, pod := range pods {
+		if pod.Name == node.Name {
+			return pod, nil
+		}
+	}
+	// Unless there is a race condition where the Pods are modified
+	// after being checked by podsEqualNodes(), this should never be reached
```
MarcelMue, line 164: I do not know if this is true.
```suggestion
	// after being checked by podsEqualNodes(), this should never be reached.
```

service/controller/resource/cleanupendpointips/create.go
```diff
@@ -149,28 +154,58 @@ func removeFromEndpointAddressList(addresses []corev1.EndpointAddress, indexesTo
...
+				cpPod, err := controlPlanePodForTCNode(node, cpPods)
+				if err != nil {
+					return len(indexesToDelete), endpoints, microerror.Mask(err)
+				}
+
+				// Check if the CP pod is Ready
+				cpPodReady := false
+				for _, c := range cpPod.Status.Conditions {
+					if c.Type == corev1.PodReady && c.Status == corev1.ConditionTrue {
+						cpPodReady = true
```
MarcelMue, line 190: this line could just be `found = true` IMO

service/controller/resource/cleanupendpointips/create.go
stone-z: When `controlPlanePodForTCNode` is called after `podsEqualNodes`, it is true because `podsEqualNodes` asserts that each Node has a same-named Pod and we don't call this function if that condition isn't met. If this were to be called in a different context, this would not necessarily be the case, but that's what the error is for.

service/controller/resource/cleanupendpointips/create.go
stone-z: Ah yeah - foggy brain. Update inbound

service/controller/resource/cleanupendpointips/create.go
```diff
@@ -149,28 +154,54 @@ func removeFromEndpointAddressList(addresses []corev1.EndpointAddress, indexesTo
 	return newAddresses
 }
 
+func controlPlanePodForTCNode(node corev1.Node, pods []corev1.Pod) (corev1.Pod, error) {
+	for _, pod := range pods {
+		if pod.Name == node.Name {
+			return pod, nil
```
tfussell, line 160: In Go, the convention is usually to make the happy path follow the left side so the last statement should be the return of the matched pod and error cases would go earlier in the function. You'd need to save the pod in a temporary variable and keep track of whether it was found or not with a bool or using a pointer.

service/controller/resource/cleanupendpointips/create.go
```diff
@@ -149,28 +154,54 @@ func removeFromEndpointAddressList(addresses []corev1.EndpointAddress, indexesTo
...
-				found = true
-				break
+				// Find the control plane pod representing this node
+				cpPod, err := controlPlanePodForTCNode(node, cpPods)
+				if err != nil {
+					return len(indexesToDelete), endpoints, microerror.Mask(err)
+				}
+
+				// Check if the CP pod is Ready
+				for _, c := range cpPod.Status.Conditions {
```
tfussell, line 187: It can make the code cleaner to move this into a `key` function. Then you could call it like `if key.PodIsReady(cpPod)`.

service/controller/resource/cleanupendpointips/create.go
```diff
@@ -149,28 +154,54 @@ func removeFromEndpointAddressList(addresses []corev1.EndpointAddress, indexesTo
...
+						found = true
+						break
+					}
+				}
+
+				// Otherwise, let this pod be removed
 			}
 		}
-		// endpoint ip does not belong to any node, lets remove it
+		// endpoint ip does not belong to any node with a healthy CP pod, lets remove it
```
tfussell, line 198:
```suggestion
		// endpoint ip does not belong to any node with a "Ready" CP pod, lets remove it
```

PR #993, opened 2020-06-30
Use tags for images, templated from operator

corest:
Towards https://github.com/giantswarm/giantswarm/issues/11859


corest:
Versions match latest SHAs.

PR #1000, opened 2020-07-07
Add docker images to watched dependencies

stone-z:


.github/dependabot.yml
```diff
@@ -1,5 +1,11 @@
 version: 2
 updates:
+- package-ecosystem: docker
```
MarcelMue, line 3: Wait do we need to set reviewers here as well?

stone-z: Ah yes, thanks. updated

PR #1006, opened 2020-07-08
update-k8s-network-setup to  latest version 0.2.0

calvix:
towards https://github.com/giantswarm/giantswarm/issues/11585

changes in `k8s-setup-network-env` are here https://github.com/giantswarm/giantswarm/issues/11585

### DIFF
--- a/service/controller/key/key.go
+++ b/service/controller/key/key.go
@@ -111,7 +111,7 @@ const (
 )
 
 const (
-	KubernetesNetworkSetupDocker = "0.1.0"
+	KubernetesNetworkSetupDocker = "0.1.0-a71a2349501386a8902c6fb2effb5048c29899a3"
 	kubernetesAPIHealthzVersion  = "0.1.0"
 )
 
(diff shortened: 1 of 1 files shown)


service/controller/key/key.go
```diff
@@ -111,7 +111,7 @@ const (
 )
 
 const (
-	KubernetesNetworkSetupDocker = "0.1.0"
+	KubernetesNetworkSetupDocker = "0.2.0"
```
MarcelMue, line 114: I am fine with this - @tfussell do you want to include this in `12.0.0` or would you rather get it in `12.0.1`?

tfussell: I reckon it can go in 12.0.0 since I'm going to try to fix the flannel bridge routing issue before releasing. I don't expect this to change anything for KVM.

calvix: Yeah this is mostly for AWS, and this change is just aligning version

But I tested cluster creation and it was fine.

PR #1015, opened 2020-07-16
Minor cleanup

MarcelMue:


MarcelMue:
> Is `/ci/default-values.yaml` not needed for linting anymore?

Turns out it is needed :D

PR #1023, opened 2020-07-27
Stop provisioning ingress NodePort Service

sslavic:
Towards https://github.com/giantswarm/giantswarm/issues/12188

### DIFF
--- a/service/controller/cloudconfig/master_template.go
+++ b/service/controller/cloudconfig/master_template.go
@@ -25,7 +25,7 @@ func (c *CloudConfig) NewMasterTemplate(customObject v1alpha1.KVMConfig, data Ig
 		params.Cluster = customObject.Spec.Cluster
 		// Ingress controller service remains in k8scloudconfig and will be
 		// removed in a later migration.
-		params.DisableIngressControllerService = false
+		params.DisableIngressControllerService = true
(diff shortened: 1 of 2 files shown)


service/controller/cloudconfig/master_template.go
```diff
@@ -25,7 +25,7 @@ func (c *CloudConfig) NewMasterTemplate(customObject v1alpha1.KVMConfig, data Ig
 		params.Cluster = customObject.Spec.Cluster
 		// Ingress controller service remains in k8scloudconfig and will be
 		// removed in a later migration.
-		params.DisableIngressControllerService = false
+		params.DisableIngressControllerService = true
```
MarcelMue, line 28: Please remove the comment as well - I assume this will go with a release where nginx becomes optional?

sslavic: @MarcelMue It's disabled for azure and now for kvm, but cleanup cannot be done (and comment is still relevant) since the same template is shared and feature still used for aws legacy.

I just created TODO for the cleanup eventually https://github.com/giantswarm/giantswarm/issues/12412

Does that make sense?

MarcelMue: Makes sense to me.

PR #1031, opened 2020-07-31
Do not push legacy operator to collection

yulianedyalkova:


stone-z:
I think you'll have to bump the architect orb version, but otherwise good to go

.circleci/config.yml
```diff
@@ -1,5 +1,5 @@
 orbs:
-  architect: giantswarm/architect@0.8.6
+  architect: giantswarm/architect@0.10.0
```
stone-z, line 2: I don't think it matters on this branch, but you could also use 0.10.1

PR #1045, opened 2020-09-18
Add gitleaks workflow

stone-z:
I'd like to test this behavior on KVM before adding it to `devctl` in https://github.com/giantswarm/devctl/pull/100

### DIFF
--- a/.github/workflows/zz_generated.gitleaks.yaml
+++ b/.github/workflows/zz_generated.gitleaks.yaml
@@ -1,0 +1,17 @@
+# DO NOT EDIT. Generated with:
+#
+#    devctl gen workflows
+#
+name: gitleaks
+
+on: [push,pull_request]
+
+jobs:
+  gitleaks:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/zz_generated.gitleaks.yaml
```diff
@@ -0,0 +1,17 @@
...
+
+jobs:
+  gitleaks:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: '0'
+    - name: gitleaks-action
+      uses: zricethezav/gitleaks-action@v1.1.4
```
tfussell, line 17: Do we need to pass any configuration to this or just run it with default settings?

stone-z: Just the defaults. If all goes well we won't need to customize it :crossed_fingers:

tfussell: Cool, let's see then.

tfussell (approved):

PR #1067, opened 2020-10-27
Update deployment on release changes

yulianedyalkova:
Towards https://github.com/giantswarm/giantswarm/issues/12480

Roll nodes if versions of calico, containerlinux, etcd or kubernetes change in the release.

This is adding a new annotation on the deployment that contains the release version and one annotation per component on the node pod with the component version last deployed.


service/controller/resource/deployment/releasecomponents.go
```diff
@@ -0,0 +1,30 @@
...
+	releasev1alpha1 "github.com/giantswarm/apiextensions/v3/pkg/apis/release/v1alpha1"
+)
+
+var keyComponents = []string{"calico", "containerlinux", "etcd", "kubernetes"}
+
+func addKeyComponentsAnnotations(annotations map[string]string, release *releasev1alpha1.Release) {
+	for _, component := range keyComponents {
+		version := componentVersion(release, component)
+		if version == "" {
+			continue
```
tfussell, line 15: Out of curiosity, why is this needed?

tfussell: Oh, `componentVersion` returns an empty string if it's not found. Nevermind.

service/controller/resource/deployment/releasecomponents.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"github.com/giantswarm/kvm-operator/service/controller/key"
+
+	releasev1alpha1 "github.com/giantswarm/apiextensions/v3/pkg/apis/release/v1alpha1"
+)
+
+var keyComponents = []string{"calico", "containerlinux", "etcd", "kubernetes"}
+
+func addKeyComponentsAnnotations(annotations map[string]string, release *releasev1alpha1.Release) {
```
tfussell, line 11: I personally would prefer a pure function here. This would require you to create a copy of the annotations map, add the component versions to the copy, and return it from the function. I'm not blocking though if you think this way is better.

service/controller/resource/deployment/releasecomponents.go
```diff
@@ -0,0 +1,30 @@
+package deployment
+
+import (
+	"github.com/giantswarm/kvm-operator/service/controller/key"
+
+	releasev1alpha1 "github.com/giantswarm/apiextensions/v3/pkg/apis/release/v1alpha1"
+)
+
+var keyComponents = []string{"calico", "containerlinux", "etcd", "kubernetes"}
```
tfussell, line 9: We usually have been calling these the "core components". Might be worth renaming this and other derivative names to keep consistency.

service/controller/key/key.go
```diff
@@ -95,6 +95,7 @@ const (
 
 const (
 	AnnotationAPIEndpoint       = "kvm-operator.giantswarm.io/api-endpoint"
+	AnnotationComponentVersion  = "kvm-operator.giantswarm.io/component-version"
```
tfussell, line 98: Since this annotation isn't used directly but is just the base name for the annotation, what do you think about calling it `AnnotationComponentVersionPrefix` or something? It would make it less surprising when we don't find it on a pod except with a suffix.

service/controller/resource/deployment/desired_test.go
```diff
@@ -306,6 +284,154 @@ func Test_Resource_Deployment_GetDesiredState(t *testing.T) {
...
+}
+
+func buildResource() (*Resource, error) {
+	// Create a fake release
+	release := releasev1alpha1.NewReleaseCR()
+	release.ObjectMeta.Name = "v1.0.0"
+	release.Spec.Components = []releasev1alpha1.ReleaseSpecComponent{
+		{
+			Name:    "kubernetes",
+			Version: "1.15.11",
```
tfussell, line 418: Might be cleaner to move these magic values like "1.15.11" into constants at the top of this file to make future changes easier and make it clear that the specific value doesn't matter, only that it matches the annotation.

service/controller/resource/deployment/desired_test.go
```diff
@@ -319,7 +445,7 @@ func Test_Resource_Deployment_GetDesiredState(t *testing.T) {
 
 		certsSearcher, err = certs.NewSearcher(c)
 		if err != nil {
-			t.Fatal("expected", nil, "got", err)
+			return nil, err
```
tfussell, line 448: We always need to mask the error when we return it. There are other spots where this should be applied.
```suggestion
			return nil, microerror.Mask(error)
```

service/controller/resource/deployment/desired_test.go
```diff
@@ -306,6 +284,154 @@ func Test_Resource_Deployment_GetDesiredState(t *testing.T) {
...
+			t.Fatalf("case %d expected %#v got %#v", i, nil, err)
+		}
+
+		deployments, ok := result.([]*v1.Deployment)
+		if !ok {
+			t.Fatalf("case %d expected %T got %T", i, []*v1.Deployment{}, result)
+		}
+
+		for _, r := range testGetComponentsAnnotations(deployments) {
+			if !reflect.DeepEqual(r, tc.ExpectedComponentsPodAnnotations) {
```
tfussell, line 400: Go maps have a random order. Does `reflect.DeepEqual` handle maps with different orders?

service/controller/resource/deployment/resource.go
```diff
@@ -107,16 +106,16 @@ func getDeploymentByName(list []*v1.Deployment, name string) (*v1.Deployment, er
...
 }
 
-func isDeploymentModified(a, b *v1.Deployment) bool {
-	aVersion, ok := a.GetAnnotations()[key.VersionBundleVersionAnnotation]
+func isAnnotationModified(a, b *v1.Deployment, annotation string) bool {
+	aVersion, ok := a.GetAnnotations()[annotation]
 	if !ok {
 		return true
 	}
 	if aVersion == "" {
```
tfussell, line 114: When `ok` is `false`, `aVersion` will be the zero value which is the empty string for `string` types. I think you can drop the `!ok` condition as a result.

service/controller/resource/deployment/resource.go
```diff
@@ -131,6 +130,18 @@ func isDeploymentModified(a, b *v1.Deployment) bool {
 	return false
```
tfussell, line 130: You should just return `aVersion != bVersion` here and drop the `if` block above.

service/controller/resource/deployment/resource.go
```diff
@@ -131,6 +130,18 @@ func isDeploymentModified(a, b *v1.Deployment) bool {
...
+func isDeploymentModified(a, b *v1.Deployment) bool {
+	if isAnnotationModified(a, b, key.VersionBundleVersionAnnotation) {
+		return true
+	}
+
+	if isAnnotationModified(a, b, key.ReleaseVersionAnnotation) {
+		return true
+	}
+
+	return false
```
tfussell, line 142:
```suggestion
	return isAnnotationModified(a, b, key.VersionBundleVersionAnnotation) || 
		isAnnotationModified(a, b, key.ReleaseVersionAnnotation)
```

tfussell:
If we have two releases, `v13.1.0` and `v13.2.0` that both use the same version of `kvm-operator` (that includes this change detection) with core components (calico, etc.) unchanged (an "app-only" upgrade), and we upgrade from `v13.1.0` to `v13.2.0`, will the nodes roll?

service/controller/resource/deployment/releasecomponents.go
yulianedyalkova: Maps are just pointers in Go so no copying is happening in this function. That's why it also doesn't also return the map, it just edits the original one. But I can still change the function if you think it's clearer. My reasoning was that it should only get the input it needs and it's easier to follow what's happening with the Annotations in the deployments.

service/controller/resource/deployment/desired_test.go
yulianedyalkova: Even in test helper functions?

service/controller/resource/deployment/desired_test.go
yulianedyalkova: It does, yes.

service/controller/resource/deployment/resource.go
yulianedyalkova: I did that somewhere in the commits and then reverted it back to how it was. Assumed we want to have an explicit check for all cases. I like what you described more too though so I'll bring it back.

service/controller/resource/deployment/resource.go
yulianedyalkova: To be honest I like the current version more. The line gets a bit too long and confusing if we merge the checks.

service/controller/resource/deployment/releasecomponents.go
yulianedyalkova: Updated 👍

service/controller/key/key.go
yulianedyalkova: Updated

service/controller/resource/deployment/desired_test.go
yulianedyalkova: Updated

service/controller/resource/deployment/resource.go
yulianedyalkova: Done!

service/controller/resource/deployment/resource.go
yulianedyalkova: Updated

yulianedyalkova:
> If we have two releases, `v13.1.0` and `v13.2.0` that both use the same version of `kvm-operator` (that includes this change detection) with core components (calico, etc.) unchanged (an "app-only" upgrade), and we upgrade from `v13.1.0` to `v13.2.0`, will the nodes roll?

It will update the deployment with the new release version in the deployment annotations but since the the pod annotations in the template spec stay the same, nodes won't get rolled.

service/controller/resource/deployment/desired_test.go
yulianedyalkova: Updated

service/controller/resource/deployment/releasecomponents.go
yulianedyalkova: Updated

service/controller/resource/deployment/resource.go
tfussell: Works for me.

service/controller/resource/deployment/desired_test.go
tfussell: It is needed for stack traces. Not sure if it's super important here, but it's good to be consistent.

service/controller/resource/deployment/resource.go
```diff
@@ -107,24 +106,26 @@ func getDeploymentByName(list []*v1.Deployment, name string) (*v1.Deployment, er
...
+	if bVersion == "" {
 		return true
 	}
-	if bVersion == "" {
+
+	return aVersion != bVersion
+}
+
+func isDeploymentModified(a, b *v1.Deployment) bool {
+	if isAnnotationModified(a, b, key.VersionBundleVersionAnnotation) {
```
tfussell, line 124: It seems to me like this will cause us to delete and recreate a deployment when the release annotation is changed (even if core components didn't change). Is this correct? See https://github.com/giantswarm/kvm-operator/blob/master/service/controller/resource/deployment/update.go#L134

tfussell: Sorry to get stuck on this, just making sure that we don't roll nodes when it's not expected. Looking at https://github.com/giantswarm/kvm-operator/blob/131ae10ae4e055650fb75840d5f3ea2597d4f25d/service/controller/resource/deployment/update.go#L33 I see that we update the Kubernetes resource when it comes from `patch.SetUpdateChange(update)` so I agree that it should be fine. I assume you have tested it as well. Let's go with this.

tfussell (approved):

PR #1068, opened 2020-10-28
Add dockerhub authentication for kubelet

wtty-fool:
Towards giantswarm/giantswarm#13981
Uses https://github.com/giantswarm/k8scloudconfig/pull/819

### DIFF
--- a/flag/service/registry/registry.go
+++ b/flag/service/registry/registry.go
@@ -3,4 +3,7 @@ package registry
 type Registry struct {
 	Domain  string
 	Mirrors string
+
+	// Registry tokens
+	DockerhubToken string
 }
--- a/helm/kvm-operator/ci/default-values.yaml
+++ b/helm/kvm-operator/ci/default-values.yaml
@@ -3,6 +3,8 @@ Installation:
   V1:
     Secret:
       Registry:
+        Dockerhub:
+          Token: "token"
         PullSecret:
           DockerConfigJSON: ""
     GiantSwarm:
--- a/helm/kvm-operator/ci/default-values.yaml
+++ b/helm/kvm-operator/ci/default-values.yaml
@@ -4,7 +4,7 @@ Installation:
     Secret:
       Registry:
         Dockerhub:
-          Token: "token"
+          Token: ""
         PullSecret:
           DockerConfigJSON: ""
     GiantSwarm:
--- a/helm/kvm-operator/templates/deployment.yaml
+++ b/helm/kvm-operator/templates/deployment.yaml
@@ -29,7 +29,7 @@ spec:
             path: config.yml
       - name: {{ .Chart.Name }}-secret
         secret:
-          name: {{ include "resource.default.name"  . }}
+          secretName: {{ include "resource.default.name"  . }}
           items:
           - key: secret.yml
             path: secret.yml
--- a/helm/kvm-operator/templates/deployment.yaml
+++ b/helm/kvm-operator/templates/deployment.yaml
@@ -27,6 +27,12 @@ spec:
           items:
           - key: config.yml
             path: config.yml
+      - name: {{ .Chart.Name }}-secret
+        secret:
+          name: {{ include "resource.default.name"  . }}
+          items:
+          - key: secret.yml
+            path: secret.yml
       serviceAccountName: {{ include "resource.default.name"  . }}
       securityContext:
         runAsUser: {{ .Values.pod.user.id }}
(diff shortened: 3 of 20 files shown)


helm/kvm-operator/ci/default-values.yaml
```diff
@@ -3,6 +3,8 @@ Installation:
   V1:
     Secret:
       Registry:
+        Dockerhub:
+          Token: "token"
```
corest, line 7: Maybe it is better to have this empty. Then we can check for empty value and fail instead of booting the node and failing to login with wrong credentials

wtty-fool: This is just for CI linting, no?

wtty-fool:
Please note `k8scloudconfig` will be merged and released first, but I'd like to get feedback ASAP.

service/service.go
```diff
@@ -153,9 +153,11 @@ func New(config Config) (*Service, error) {
 				GroupsClaim:    config.Viper.GetString(config.Flag.Service.Installation.Tenant.Kubernetes.API.Auth.Provider.OIDC.GroupsClaim),
 				GroupsPrefix:   config.Viper.GetString(config.Flag.Service.Installation.Tenant.Kubernetes.API.Auth.Provider.OIDC.GroupsPrefix),
 			},
+			SSOPublicKey: config.Viper.GetString(config.Flag.Service.Tenant.SSH.SSOPublicKey),
+
+			DockerhubToken:  config.Viper.GetString(config.Flag.Service.Registry.DockerhubToken),
```
tfussell, line 158: Is this something specific to Docker Hub or is this just normal docker credentials? What happens when the registry domain is Quay (or something other than Docker)?

wtty-fool: It's docker credentials used by `k8s-kubelet.service` to authorize/authenticate against DockerHub API.
When they are missing, we might hit QPS limits.
If registry domain is set to anything other than DockerHub, they are not utilized, but still present on the node fs.

tfussell: It feels like `DockerRegistryCredentials` would have been a better name since it's not Hub specific. But this name is used in the other PRs so it's fine for now.

tfussell (approved):
LGTM if tested.

flag/service/registry/registry.go
```diff
@@ -3,4 +3,7 @@ package registry
 type Registry struct {
 	Domain  string
 	Mirrors string
+
+	// Registry tokens
+	DockerhubToken string
```
kopiczko, line 8: Shouldn't we split it here to `Dockerhub.Token` too?

wtty-fool: Let's not.

helm/kvm-operator/templates/deployment.yaml
```diff
@@ -36,11 +42,13 @@ spec:
         image: "{{ .Values.Installation.V1.Registry.Domain }}/{{ .Values.image.name }}:{{ .Values.image.tag }}"
         args:
         - daemon
-        - --config.dirs=/var/run/{{ .Chart.Name }}/configmap/
-        - --config.files=config
+        - --config.dirs=/var/run/{{ .Chart.Name }}/configmap/,/var/run/{{ .Chart.Name }}/secret/
```
kopiczko, line 45: This is how it's done in aws-operator:

```
        - --config.dirs=/var/run/{{ include "name" . }}/configmap/
        - --config.dirs=/var/run/{{ include "name" . }}/secret/
```

helm/kvm-operator/templates/secret.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: Secret
+type: Opaque
+metadata:
+  name: {{ include "resource.default.name" . }}
+  namespace: {{ include "resource.default.namespace" . }}
+  labels:
+    {{- include "labels.common" . | nindent 4 }}
+data:
+  secret.yml: {{ printf "service: {registry: {dockerhubToken: %q}}" .Values.Installation.V1.Secret.Registry.Dockerhub.Token | b64enc | quote }}
```
kopiczko, line 10: @corest shouldn't we have something like this here?

```
  secret.yaml: {{ .Values.Installation.V1.Secret.KVMOperator.SecretYaml | b64enc | quote }}
```

kopiczko: Talked with @corest. That should be a separate file. To be on par with azure-operator it could be named `dockerhub-secret.yaml:`

kopiczko (approved):

PR #1082, opened 2020-11-16
Use TC node IP from status and not labels

yulianedyalkova:
Towards https://github.com/giantswarm/giantswarm/issues/13887

### DIFF
--- a/service/controller/resource/cleanupendpointips/create.go
+++ b/service/controller/resource/cleanupendpointips/create.go
@@ -179,7 +179,11 @@ func removeDeadIPFromEndpoints(endpoints *corev1.Endpoints, nodes []corev1.Node,
 		found := false
 		// check if the ip belongs to any k8s node
 		for _, node := range nodes {
-			if node.Labels["ip"] == ip.IP {
+			nodeIP, err := key.NodeInternalIP(node)
+			if err != nil {
+				return len(indexesToDelete), endpoints, microerror.Mask(err)
+			}
+			if nodeIP == ip.IP {
 				// Find the control plane pod representing this node
 				cpPod, err := controlPlanePodForTCNode(node, cpPods)
 				if err != nil {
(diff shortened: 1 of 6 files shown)


service/controller/resource/cleanupendpointips/create.go
```diff
@@ -179,7 +179,11 @@ func removeDeadIPFromEndpoints(endpoints *corev1.Endpoints, nodes []corev1.Node,
 		found := false
 		// check if the ip belongs to any k8s node
 		for _, node := range nodes {
-			if node.Labels["ip"] == ip.IP {
+			nodeIP, err := key.NodeInternalIP(node)
```
MarcelMue, line 182: At which point is the IP written to the status? Is there a different order of execution now or has everything essentially stayed the same?

I am concerned that we might introduce a hidden dependency on another controller (which likely already exists) but are not certain of the timings.

yulianedyalkova: When we start kubelet, we pass the IP as  `--node-ip` and `--node-labels` so I assumed that kubelet is somehow responsible for populating both. But I'll dig deeper into that.

MarcelMue: Sorry I wasn't clear:
I am not concerned about the node IP on the k8s object but the status vs label thing

yulianedyalkova: Setting the node status is one of the first things that kubelet does once it starts running ([click](https://github.com/kubernetes/kubernetes/blob/1df1f882c4421d313abb468fef5c15f89983471c/pkg/kubelet/kubelet.go#L1394)). It iterates and calls a bunch of setters and [NodeStatus](https://github.com/kubernetes/kubernetes/blob/1df1f882c4421d313abb468fef5c15f89983471c/pkg/kubelet/nodestatus/setters.go#L60) is one of them. I haven't checked who sets the labels but I'll be surprised if it's happening earlier. I also tested this on `gorgoth` and it sounded happy.

MarcelMue (approved):
I am convinced :D

tfussell (approved):
LGTM if tested.

PR #1092, opened 2020-12-04
Using new micrologger 0.4.0 

tomahawk28:
Introducing new micrologger functions so make logs simpler.


tfussell (approved):
LGTM. Hope you used a find and replace for that!

PR #1103, opened 2020-12-09
Skip drain on deletion

yulianedyalkova:
Towards https://github.com/giantswarm/giantswarm/issues/14291


stone-z (approved):
Seems reasonable to me :+1:

service/controller/resource/pod/delete.go
```diff
@@ -41,6 +41,28 @@ func (r *Resource) EnsureDeleted(ctx context.Context, obj interface{}) error {
...
+		}
+
+		kvmConfig, err := r.g8sClient.ProviderV1alpha1().KVMConfigs(metav1.NamespaceDefault).Get(ctx, clusterID, metav1.GetOptions{})
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		if kvmConfig.DeletionTimestamp != nil {
+			r.logger.LogCtx(ctx, "level", "debug", "message", "cluster is being deleted")
+			resourcecanceledcontext.SetCanceled(ctx)
```
MarcelMue, line 59: Is this necessary? I would have thought that just returning `nil` would do the trick here. There is no other action following here in this resource or am I missing something?

MarcelMue (approved):
Fine with me - just one possible improvement.

service/controller/resource/pod/delete.go
yulianedyalkova: We're [using](https://github.com/giantswarm/kvm-operator/blob/682a07b57f78d1742c27c2c42a858b7eefd77e51/service/controller/resource/pod/delete.go#L50-L52) this all across the file so I just kept things consistent. I don't know if it does anything though - we're calling `SetCanceled` in a lot of places in the operator but no one calls `IsCanceled`.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Do not drain node pods when cluster is being deleted
```
tfussell, line 12:
```suggestion
- Do not drain node pods when cluster is being deleted.
```

tfussell (approved):

PR #1105, opened 2020-12-09
Update k8scc to v10

tfussell:


tfussell:
Tested on geckon with k8s 1.19.

MarcelMue (approved):

PR #1106, opened 2020-12-10
Add unhealthy node detector controller

yulianedyalkova:
Towards https://github.com/giantswarm/roadmap/issues/176

Auto terminate NotReady TC nodes
* manually add `alpha.node.giantswarm.io/terminate-unhealthy` annotion on  the`KVMConfig` for a cluster ([this](https://docs.giantswarm.io/basics/automatic-termination-of-bad-nodes/) is how we currently do it for other providers as well)
* watch all TC nodes for clustes with the annotation
* if the node is `NotReady` for 6 reconcile loops -> delete the node pod from CP


service/controller/resource/terminateunhealthynodes/create.go
```diff
@@ -0,0 +1,95 @@
...
+		}
+	}
+
+	var detectorService *detector.Detector
+	{
+		detectorConfig := detector.Config{
+			K8sClient: tcCtrlClient,
+			Logger:    r.logger,
+
+			NotReadyTickThreshold: nodeTerminationTickThreshold,
```
tfussell, line 46: We should probably make this an absolute time rather than a number relative to the reconciliation frequency in the future in case we ever change the period away from 5 minutes. Fine for now though.

service/controller/resource/terminateunhealthynodes/create.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+func (r *Resource) terminateNode(ctx context.Context, clusterID string, node corev1.Node) error {
+	r.logger.Debugf(ctx, "getting corresponding CP pod for node %s", node.Name)
+	pod, err := r.k8sClient.CoreV1().Pods(clusterID).Get(ctx, node.Name, metav1.GetOptions{})
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	r.logger.Debugf(ctx, "terminating unhealhty node %s", node.Name)
```
MarcelMue, line 86:
```suggestion
	r.logger.Debugf(ctx, "terminating unhealthy node %s", node.Name)
```

service/controller/resource/terminateunhealthynodes/create.go
MarcelMue: I even think this timing might currently be different between AWS and KVM with this - I thought AWS reconciled a bit quicker than 5 minutes. Maybe I am wrong though.

A time would be better though, I am not sure how this reacts if you edit the CR a bunch and force it to reconcile quicker that way?

tfussell: Good point. If we changed the label on a KVMConfig 6 times in a few seconds, for example, would we use up all 6 ticks and terminate the node too quickly? Unlikely, but worth considering.

calvix: in aws-operator and azure operator the  node termination logic is in the separate controller so does not depend on the reconciliation of the cluster controller, and the reconciliation loop is set to 2 mins

The absolute time was discouraged by several ppl from sig operator because we work with an event-based system and that can be problematic, especially if you work with just time than flapping node which is ready for 10sec out of 5 mins can be overlooked without the tick logic.

Also, time can be skewed on different servers due to some errors like NTP or so, we know cases where cronjobs in China are not started cause of absolute time issues.

yulianedyalkova: The detector here runs in a separate controller too and it uses the default resync period (5m). So we can change the period for other controllers independently.

I just tried editing the kvmconfig while a node is not ready and you are right - it does bump the ticker count. Is there a way to tell the controller to only reconcile after the timeout and not on changes? 

This shouldn't really be a problem right now though because we write the kvmconfig once and we don't fiddle with it a lot after that (correct me if I'm wrong). And it bumps the ticker only if the node is already in a NotReady state so it won't affect happy nodes. I was planning to change the detector to use a timestamp but @calvix has a really good point about the different servers.

tfussell: True, we have run into clock skew problems in the past resulting in unexpected things breaking (like Prometheus). I suppose you could cache the last reconciliation time of the KVMConfig and skip reconciliation if the last reconciliation time is less than the reconciliation period. Go uses a monotonic clock in addition to the wall clock so even if the wall clock changes during execution, I don't think the reconciliation would be affected. When we restart the operator, we would lose this information, but this would use up at most one tick before the values are cached again.

calvix: The absolute time is also tricky because if you just compare for how long the node is not ready then you are going miss cases when a node is flapping ( is ready for 30 sec and not ready for 3 mins) as from the transition time on the node it will look fine. 

example:

node is flapping and is `ready` only 30s out of 3 minutes, in the transition time it will always look that the node is  in a not ready state for a max of 2min 30 sec ( so not for termination)

but with the tick systems, it would eventually reach the threshold for termination as it does not compare the absolute time but rather look at series of even when node was seen often as not ready

calvix: The library works now how it works and was discussed with several ppl and everyone was fine with it, if team rocket needs custom implementation then you are free to do so, I am not keen to rebuild the lib for the third time. If you feel it's  bad and stupid then we need a bigger discussion probably across sig infra snd sig operator to make a decision.

I am stepping out of this for now :)

yulianedyalkova: I would also like to merge this PR to keep up with other provider operators. We can always reiterate and improve future versions of the detector if we feel the need to.

tfussell (approved):
This matches the implementation on the other providers so I'm fine with it for now. We'll improve it later if needed.

PR #1127, opened 2021-01-28
Increase memory limit and request

tfussell:
In certain deployments we have `kvm-operator` reaching 300MB of memory usage causing it to be killed. 400MB gives us ~33% of wiggle room.

PM here https://github.com/giantswarm/giantswarm/issues/15101

### DIFF
--- a/helm/kvm-operator/templates/deployment.yaml
+++ b/helm/kvm-operator/templates/deployment.yaml
@@ -58,9 +58,9 @@ spec:
         resources:
           requests:
             cpu: 250m
-            memory: 250Mi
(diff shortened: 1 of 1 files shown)


mazzy89 (approved):

tfussell:
Waiting to test an approach for reducing memory usage before merging this https://github.com/giantswarm/operatorkit/pull/512

yulianedyalkova:
Also as discussed, I'm not sure this will have any effect now that we're using VPA.

tfussell:
Considering the comments above, I'm going to close this and pursue other approaches.

PR #1128, opened 2021-01-28
Remove IPs from not ready WC nodes

tfussell:
Towards https://github.com/giantswarm/giantswarm/issues/15266

PR #1132, opened 2021-02-02
Add per-WC controller to watch node ready state

tfussell:
More elegant alternative to https://github.com/giantswarm/kvm-operator/pull/1128 for https://github.com/giantswarm/giantswarm/issues/13285

It's possible for a Kubernetes node in a WC to become NotReady while it still appears healthy at the VM level (and therefore from the MC), for example if kubelet is not running on the node. Even worse, if CNI isn't ready then the node is NotReady but kubelet is responding and we couldn't tell without requesting the status from the WC kubernetes API. An alternative would be to list nodes on each reconciliation loop and remove a node's endpoint IP if the node becomes NotReady, but it can take several minutes which is too long for traffic to be dropped to a node. By creating a controller dynamically for each WC, we can subscribe to events related to the nodes in the WC and update MC Endpoints when node readiness changes almost immediately.


go.mod
```diff
@@ -13,7 +13,7 @@ require (
 	github.com/giantswarm/microerror v0.3.0
 	github.com/giantswarm/microkit v0.2.2
 	github.com/giantswarm/micrologger v0.5.0
-	github.com/giantswarm/operatorkit/v4 v4.2.0
+	github.com/giantswarm/operatorkit/v4 v4.2.1-0.20210204214713-f1a5d6e536d3
```
tfussell, line 16: Will use released version once https://github.com/giantswarm/operatorkit/pull/516 is merged

service/controller/deleter_resource_set.go
```diff
@@ -6,26 +6,12 @@ import (
...
 	"github.com/giantswarm/operatorkit/v4/pkg/resource/wrapper/retryresource"
 
-	"github.com/giantswarm/kvm-operator/service/controller/resource/cleanupendpointips"
 	"github.com/giantswarm/kvm-operator/service/controller/resource/node"
 )
 
 func newDeleterResources(config DeleterConfig) ([]resource.Interface, error) {
 	var err error
 
-	var cleanupendpointipsResource resource.Interface
```
tfussell, line 16: We don't need this resource any more because the new nodecontroller and the extra logic in the pod resource handler totally subsume what this did in a better way.

service/controller/drainer.go
```diff
@@ -48,7 +47,6 @@ func NewDrainer(config DrainerConfig) (*Drainer, error) {
 			Logger:    config.Logger,
 			Resources: resources,
 			Selector: labels.SelectorFromSet(map[string]string{
-				key.PodWatcherLabel:   project.Name(),
```
tfussell, line 51: I don't see a use for this. Let me know if I'm missing anything.

service/controller/internal/nodecontroller/controller.go
```diff
@@ -0,0 +1,101 @@
+package nodecontroller
```
tfussell, line 1: This is a dynamically created/removed controller which exists for each KVMConfig resource and which watches a WC API for node events (mostly Ready condition).

service/controller/internal/nodecontroller/resource/readylabel/create.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"context"
+
+	"github.com/giantswarm/microerror"
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/giantswarm/kvm-operator/service/controller/key"
+)
+
+func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
```
tfussell, line 12: This is the only resource in the dynamic controller. It manages an annotation to MC node pods reflecting the ready status of corresponding WC nodes.

service/controller/internal/nodecontroller/resource/readylabel/create.go
```diff
@@ -0,0 +1,45 @@
+package readylabel
```
tfussell, line 1: I just realized this should be called `readyannotation`. I'll rename it tomorrow.

service/controller/internal/nodecontroller/resource/readylabel/create.go
```diff
@@ -0,0 +1,45 @@
...
+	patch := annotationPatch{
+		key:   key.AnnotationPodNodeStatus,
+		value: key.PodNodeStatusReady,
+	}
+	if !nodeReady {
+		patch.value = key.PodNodeStatusNotReady
+	}
+
+	var managementPod corev1.Pod
+	err := r.managementK8sClient.Get(ctx, key.NodePodKey(r.cluster, *workloadNode), &managementPod)
```
tfussell, line 35: Might be able to skip getting the pod before patching it.

service/controller/key/key.go
```diff
@@ -203,6 +200,13 @@ func ClusterNamespace(customObject v1alpha1.KVMConfig) string {
 	return ClusterID(customObject)
 }
 
+func NodePodKey(cluster v1alpha1.KVMConfig, node corev1.Node) client.ObjectKey {
+	return client.ObjectKey{
```
tfussell, line 204: Used with the `ctrlClient` to get the key of the node pod corresponding to a given cluster and WC node.

service/controller/key/key.go
```diff
@@ -234,38 +238,13 @@ func CPUQuantity(n v1alpha1.KVMConfigSpecKVMNode) (resource.Quantity, error) {
...
-	customObject, err := ToCustomObject(obj)
-	if err != nil {
-		return nil, microerror.Mask(err)
-	}
-
-	i := ClusterID(customObject)
-	e := ClusterAPIEndpoint(customObject)
+// CreateK8sClientForWorkloadCluster takes the context of the reconciled object
+// and the provided logger and tenant cluster interface and creates a K8s client for the workload cluster
+func CreateK8sClientForWorkloadCluster(ctx context.Context, cluster v1alpha1.KVMConfig, logger micrologger.Logger, workloadCluster tenantcluster.Interface) (*k8sclient.Clients, error) {
```
tfussell, line 243: Simplified since we want the k8sclient rather than just the kubernetes client.

service/controller/resource/endpoint/desired.go
```diff
@@ -35,6 +35,11 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
 		}
 	}
 
+	statusAnnotation, hasStatusAnnotation := pod.Annotations[key.AnnotationPodNodeStatus]
+	if hasStatusAnnotation && statusAnnotation != "" {
```
tfussell, line 39: Only consider the annotation if it's present (in case the annotation controller isn't working for some reason).

service/controller/resource/nodecontroller/create.go
```diff
@@ -0,0 +1,84 @@
...
+
+func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
+	cr, err := key.ToCustomObject(obj)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	r.logger.Debugf(ctx, "ensuring node controller is running for workload cluster")
+
+	if _, ok := r.controllers[nodeControllerKey(cr)]; !ok {
```
tfussell, line 24: I'm not sure if I should recreate the controller on every reconciliation. I'm worried that certs, for example, could change and the controller would be silently failing.

service/controller/resource/pod/create.go
```diff
@@ -2,8 +2,119 @@ package pod
...
+
+	"github.com/giantswarm/microerror"
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/giantswarm/kvm-operator/service/controller/key"
 )
 
 func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
+	pod, err := key.ToPod(obj)
```
tfussell, line 14: This logic comes from `cleanupendpointips` but is now pod-specific rather than for the whole cluster. It should be more efficient.

.stignore
```diff
@@ -1,18 +1,2 @@
...
-
-# Test binary, built with go test -c
-*.test
-
-# Output of the go coverage tool, specifically when used with LiteIDE
-*.out
-
-# dlv binary
-__debug_bin
+kvm-operator
```
MarcelMue, line 2: nit: Missing newline

service/controller/drainer.go
MarcelMue: I think the risk here is that you might select pods which coincidentally have the operator version label.

service/controller/resource/nodecontroller/create.go
MarcelMue: I am a bit worried about dynamically creating these controllers. From the current logic implemented it looks reasonably safe. I could see it causing accidental memory leaks easily though.

Booting the controllers on each reconciliation could create an unreasonable amount of logs.

service/controller/drainer.go
tfussell: That makes a lot of sense. Unlikely, but doesn't hurt to keep it.

service/controller/resource/nodecontroller/create.go
tfussell: I have added some very careful code to ensure that we would only reboot a controller if the certs or KVMConfig spec changes and to make sure that only one controller per cluster can run at one time. Let me know what you think of the logic if you have time to check it out.

tfussell: Also I'll take a look at the logs to make sure it's still readable.

service/controller/resource/nodecontroller/create.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+// applyCreateChangeAsync boots the desired controller and saves it in the controllers map
+// if it boots successfully. It should be run as a goroutine to avoid blocking reconciliation of
+// other resources.
+func (r *Resource) applyCreateChangeAsync(ctx context.Context, desired controllerWithConfig) {
+	r.logger.Debugf(ctx, "DELETEME before boot")
+	desired.Boot(ctx)
+	r.logger.Debugf(ctx, "DELETEME after boot")
+	<-desired.Booted() // Wait for the controller to be booted
```
MarcelMue, line 46: Is there a time limit here? Do we expect this function to always return in a reasonable timeframe?

MarcelMue (approved):
In general this is LGTM. I would appreciate more reviews.

service/controller/resource/nodecontroller/create.go
tfussell: Internally it has a retry back off. If the retry fails a few times, it will exit the whole program with `os.Exit` we don't have to worry about this getting stuck permanently.

tfussell: Looks like it tries 7 times with a 1 second linear backoff, so no more than 7 seconds https://github.com/giantswarm/operatorkit/blob/07b70cc9b37090b1c102bcfa35e78f54edffce5c/pkg/controller/controller.go#L236

MarcelMue: I see - that's a reasonable timeframe.

service/controller/drainer.go
```diff
@@ -9,9 +9,10 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 
+	"github.com/giantswarm/kvm-operator/service/controller/key"
```
yulianedyalkova, line 12: Is this intentional?

tfussell: That looks strange. Goimports should have caught that. Let me see.

tfussell: Fixed.

service/controller/key/key.go
```diff
@@ -474,16 +446,24 @@ func NetworkDNSBlock(servers []net.IP) string {
...
 
-func NetworkNTPBlock(servers []net.IP) string {
-	var ntpBlockParts []string
-
-	for _, s := range servers {
-		ntpBlockParts = append(ntpBlockParts, fmt.Sprintf("NTP=%s", s.String()))
+// FindNodeCondition returns the condition of the given type from the node. The second return value indicates if the condition was found.
+func FindNodeCondition(node corev1.Node, conditionType corev1.NodeConditionType) (corev1.NodeCondition, bool) {
+	for _, condition := range node.Status.Conditions {
+		if condition.Type == corev1.NodeReady {
```
yulianedyalkova, line 452: `conditionType` is not being used in the fuction. I'm assuming you want to check for it here?

service/controller/key/key.go
```diff
@@ -474,16 +446,24 @@ func NetworkDNSBlock(servers []net.IP) string {
...
 	}
+	return corev1.NodeCondition{}, false
+}
 
-	ntpBlock := strings.Join(ntpBlockParts, "\n")
-
-	return ntpBlock
+// FindNodeCondition returns the condition of the given type from the pod. The second return value indicates if the condition was found.
+func FindPodCondition(pod corev1.Pod, conditionType corev1.PodConditionType) (corev1.PodCondition, bool) {
+	for _, condition := range pod.Status.Conditions {
```
yulianedyalkova, line 461: `conditionType` is not being used in the function. I'm assuming you want to check for it here?

service/controller/internal/nodecontroller/resource/podcondition/create.go
```diff
@@ -0,0 +1,66 @@
...
+
+	return nil
+}
+
+func calculatePodNodeCondition(node corev1.Node, pod corev1.Pod) (corev1.PodCondition, bool) {
+	nodeReadyCondition, nodeReadyFound := key.FindNodeCondition(node, corev1.NodeReady)
+	podReadyCondition, podReadyFound := key.FindPodCondition(pod, key.WorkloadClusterNodeReady)
+
+	condition := podReadyCondition
+	condition.Type = key.WorkloadClusterNodeReady // In case FindPodCondition returned an empty object
```
yulianedyalkova, line 44: Currently you're always overriding the value, maybe add a check for `podReadyFound` around this?

service/controller/internal/nodecontroller/resource/podcondition/create.go
```diff
@@ -0,0 +1,66 @@
...
+	condition.Message = nodeReadyCondition.Message
+	condition.Reason = nodeReadyCondition.Reason
+
+	if nodeReadyFound {
+		if nodeReadyCondition.Status == corev1.ConditionTrue {
```
yulianedyalkova, line 49: I would personally rewrite this if to something like this:
```
	switch nodeReadyCondition.Status {
	case corev1.ConditionTrue:
		condition.Status = corev1.ConditionTrue
	case corev1.ConditionFalse:
		condition.Status = corev1.ConditionFalse
	default:
		condition.Status = corev1.ConditionUnknown
	}
```
or this:
```
	condition.Status = corev1.ConditionUnknown
	if nodeReadyCondition.Status == corev1.ConditionTrue || nodeReadyCondition.Status == corev1.ConditionFalse {
		condition.Status = nodeReadyCondition.Status
	}
```

But nothing critical.

tfussell: I'll turn it into a switch.

service/controller/key/key.go
tfussell: Good catch. I'll fix it.

service/controller/internal/nodecontroller/resource/podcondition/create.go
yulianedyalkova: Or even this
```
	switch nodeReadyCondition.Status {
	case corev1.ConditionTrue, corev1.ConditionFalse:
		condition.Status = nodeReadyCondition.Status
	default:
		condition.Status = corev1.ConditionUnknown
	}
```

service/controller/internal/nodecontroller/controller.go
```diff
@@ -0,0 +1,106 @@
...
+		c := controller.Config{
+			K8sClient: config.WorkloadK8sClient,
+			Logger:    config.Logger,
+			Resources: resources,
+			NewRuntimeObjectFunc: func() runtime.Object {
+				return new(corev1.Node)
+			},
+			Selector: labels.SelectorFromSet(map[string]string{
+				label.OperatorVersion: project.Version(),
+				"role":                key.WorkerID,
```
yulianedyalkova, line 54: Why only worker nodes?

service/controller/resource/nodecontroller/create.go
```diff
@@ -0,0 +1,156 @@
...
+		cluster:    cluster,
+		restConfig: restConfig,
+		Controller: nodeController,
+	}
+	return withConfig, nil
+}
+
+// controllersEqual returns true when the given controllers are equal as it relates to watching the workload
+// cluster Kubernetes APIs.
+func (r *Resource) controllersEqual(ctx context.Context, left controllerWithConfig, right controllerWithConfig) bool {
```
stone-z, line 133: I'm wondering if this logic (and its helpers) should belong to `controllerWithConfig`, either statically or as an instance method (e.g. `left.Equals(right)`). That way if its API changes this implementation is more likely to follow.

service/controller/internal/nodecontroller/controller.go
tfussell: We only manage endpoint IPs for workers. We don't remove the master IP even if the node or pod is down because then we wouldn't be able to access the Kubernetes API to check status and add it back once it's healthy again (chicken-and-egg problem).

stone-z (approved):
Included some thoughts but I like your controller-per-WC idea. This is cool

service/controller/key/key.go
```diff
@@ -525,13 +505,23 @@ func OperatorVersion(cr v1alpha1.KVMConfig) string {
...
+			return true
 		}
 	}
-	return podReady
+	return false
+}
+
+// PodNodeIsReady examines the Status Conditions of a Pod
+// and returns true if the WorkloadClusterNodeReady Condition is true.
+func PodNodeIsReady(pod corev1.Pod) bool {
```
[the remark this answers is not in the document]
tfussell: Just did :)

service/controller/internal/nodecontroller/controller.go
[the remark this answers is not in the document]
stone-z: I see Yulia had the same question in the time it took me to finish. I'd suggest adding a note there since we both noticed.
As I think about it, I guess we can't currently remove the master endpoint, but there is still the problem that workloads running on the master will continue to get routed traffic which will fail. I wonder if in the future we can either not rely solely on the Service endpoints ourselves, somehow drop only workload endpoints, or if this is just something we'll have to live with.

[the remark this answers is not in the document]
tfussell: I'll add a comment about why it's watching workers only.

I think as long as the master exists we should continue to route traffic to it even if it's not currently healthy. This might change once we move to multi-master on KVM as you say.

service/controller/internal/nodecontroller/resource/podcondition/create.go
tfussell: Good improvement. Done.

service/controller/resource/nodecontroller/create.go
tfussell: I agree that's it's better to keep them close. I made it into a method on the resource handler so we could have the logger, but those lines are just for my debugging so I'll drop them and move the function to be a member of `controllerWithConfig`.

yulianedyalkova (approved):

service/controller/internal/nodecontroller/controller.go
```diff
@@ -0,0 +1,108 @@
...
+		c := controller.Config{
+			K8sClient: config.WorkloadK8sClient,
+			Logger:    config.Logger,
+			Resources: resources,
+			NewRuntimeObjectFunc: func() runtime.Object {
+				return new(corev1.Node)
+			},
+			Selector: labels.SelectorFromSet(map[string]string{
+				label.OperatorVersion: project.Version(),
+				// When managing endpoints, e only consider node- and pod-readiness for workers. Master endpoint IPs
```
stone-z, line 54:
```suggestion
				// When managing endpoints, we only consider node- and pod-readiness for workers. Master endpoint IPs
```

service/controller/resource/nodecontroller/create.go
tfussell: Fixed.

tfussell:
I found in my testing today that upgrades from this operator to the same operator couldn't complete because the endpoint was removed from the master before it was drained and everything got stuck. This is fixed now.

The last problem I see here is that there are spurious `operatorkit` errors while the master is down during an upgrade coming from `controller-runtime` and the timestamp collector. This isn't a showstopper, but it means we're going to have increased error rates (possibly triggering alerts) and more noise in logs. For the collector error, I would make the prometheus metrics able to be disabled (we don't care about WC node timestamps). For the controller-runtime errors, I might try to disable the watch when the WC API is not available.

tfussell:
This is already getting quite long and it's hard to review when there are new changes after already approving. I'm going to merge now and open a new PR for the remaining issues.

tfussell:
This is already getting quite long and it's hard to review when there are new changes after already approving. I'm going to merge now and open a new PR for the remaining issues.

tfussell:
This is already getting quite long and it's hard to review when there are new changes after already approving. I'm going to merge now and open a new PR for the remaining issues.

PR #1136, opened 2021-02-08
Test operatorkit timestamp collector improvements

tfussell:


tfussell:
@yulianedyalkova I merged master in here to get the latest endpoint IP fixes. I don't think the collector/operator stuff should be affected.

yulianedyalkova:
So this is the comparison between master/wip branch:

master:
Screenshot 2021-02-24 at 10 32 33

wip:
Screenshot 2021-02-24 at 10 57 22

The only difference is that in the second case, the `kind` label is missing. I tested it and it turns out that the object that we get back does not have the [kind](https://github.com/giantswarm/operatorkit/blob/0f90ab2421793a74e4d0f362966c144317918fa3/pkg/controller/collector/timestamp.go#L91) field.

Switching back to to an `UnstructuredList` instead of `PartialObjectMetadataList` resolved the issue.

tfussell:
No longer needed.

PR #1153, opened 2021-02-17
Rename tenant to workload

tfussell:
We recently decided to follow Cluster API conventions and call our tenant clusters workload clusters and our control planes management clusters. This isn't strictly necessary in our code, but I hope it will reduce mental burden when making future changes and improve consistency.

I also removed two unused flags that I noticed.


helm/kvm-operator/templates/configmap.yaml
```diff
@@ -33,7 +33,7 @@ data:
         ntp:
           servers: {{ .Values.Installation.V1.Provider.KVM.NTP.Servers }}
         {{- if .Values.Installation.V1.Guest }}
-        tenant:
+        workload:
```
tfussell, line 36: This is the only change I could conceivably see having an impact. As far as I can tell, we don't override this in `installations` for any customer though.

yulianedyalkova (approved):

flag/service/workload/tenant.go
```diff
@@ -0,0 +1,11 @@
+package workload
```
stone-z, line 1: Maybe this file should also be named `workload.go`

MarcelMue (approved):
Seems fine. The change in flags should be tested.

stone-z (approved):
Sounds good

flag/service/workload/tenant.go
tfussell: Thanks, I searched in files but didn't look at filenames. This is fixed now.

MarcelMue:
Is there a blocker to merge this?

tfussell:
Tested running on `gorgoth`. No issues with flag renaming. Finally merging.

PR #1161, opened 2021-02-24
Use controller-runtime client throughout

tfussell:
This PR affects many files, but the actual change is relatively small. We are moving towards using the controller-runtime client instead of the typed clients `G8sClient` and `K8sClient` in our operators. Instead of slowly migrating one resource handler at a time, I thought it would make our code easier to read if we standardize the client everywhere. I tested this when I first wrote it, but will test it again once CircleCI is working.


tfussell:
If we continue work on KVM, it will be a new operator so I'm closing this.

PR #1198, opened 2021-05-04
Improve requeue logic

tfussell:
I missed a requeue on successful reconciliation of the WC nodes. This meant that the operator would recreate the WC watcher because it looked like it was failed. Also improving naming of requeues and an error message.


service/controller/resource/nodecontroller/create.go
```diff
@@ -73,7 +73,7 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 			r.logger.Debugf(ctx, "controllers don't match")
 			shouldStop = true
 		} else if time.Since(current.LastReconciled()) > 2*nodecontroller.ResyncPeriod {
-			r.logger.Debugf(ctx, "controller hasn't reconciled in more than %d minutes", nodecontroller.ResyncPeriod.Minutes())
+			r.logger.Debugf(ctx, "controller hasn't reconciled in more than %d minutes", int(2*nodecontroller.ResyncPeriod.Minutes()))
```
pipo02mix, line 76: why multiple by 2?

pipo02mix (approved):
LGTM just a question

service/controller/resource/nodecontroller/create.go
tfussell: There is some variability in reconciliation duration. If we checked after exactly 5 minutes, it might or might not have completed. If there's not a reconciliation after 10 minutes, then we are guaranteed to have missed at least one reconciliation.

PR #1208, opened 2021-05-19
Add host volumes support

mazzy89:
Add host volumes support:

- Add the env var `HOST_DATA_VOLUME_CONFIG` to the `k8s-kvm` container
- Add two mount units to configure the volumes

### DIFF
--- a/service/controller/cloudconfig/worker_template.go
+++ b/service/controller/cloudconfig/worker_template.go
@@ -46,7 +46,6 @@ func (c *CloudConfig) NewWorkerTemplate(ctx context.Context, cr v1alpha1.KVMConf
 		params.KVMWorkerMountTags = []string{}
 
 		if len(cr.Spec.KVM.Workers) > 0 {
-
 			for _, hostVolume := range cr.Spec.KVM.Workers[nodeIndex].HostVolumes {
 				params.KVMWorkerMountTags = append(params.KVMWorkerMountTags, hostVolume.MountTag)
 			}


service/controller/resource/deployment/worker_deployment.go
```diff
@@ -42,6 +42,12 @@ func newWorkerDeployments(customResource v1alpha1.KVMConfig, release *releasev1a
 			return nil, microerror.Maskf(invalidConfigError, "error creating memory quantity: %s", err)
 		}
 
+		for _, hostVolume := range capabilities.HostVolumes {
+			if hostVolume.MountTag == "" || hostVolume.HostPath == "" {
```
tfussell, line 46: Should we push this validation into the CRD schema to make sure an invalid CR can't even be created? Could use a regex or set `MinLength` to 1.

service/controller/resource/deployment/worker_deployment.go
```diff
@@ -383,9 +389,19 @@ func newWorkerDeployments(customResource v1alpha1.KVMConfig, release *releasev1a
...
 
 	return deployments, nil
 }
+
+func addConditionalEnvVarsToK8SKVMContainer(deployment *v1.Deployment, envVars []corev1.EnvVar) {
+	for _, container := range deployment.Spec.Template.Spec.Containers {
+		if container.Name == key.K8SKVMContainerName {
+			container.Env = append(container.Env, envVars...)
+		}
+	}
```
tfussell, line 406: `container` is a value rather than a reference in the loop so I think you'll need to do something like this to get it to change the passed in deployment:
```suggestion
	for i, container := range deployment.Spec.Template.Spec.Containers {
		if container.Name == key.K8SKVMContainerName {
			container.Env = append(container.Env, envVars...)
			deployment.Spec.Template.Spec.Containers[i] = container
		}
	}
```

service/controller/key/key.go
```diff
@@ -348,6 +350,25 @@ func KubeletVolumeSizeFromNode(node v1alpha1.KVMConfigSpecKVMNode) string {
...
+func HostVolumesToEnvVar(hostVolumes []v1alpha1.KVMConfigSpecKVMNodeHostVolumes) corev1.EnvVar {
+	var lastElemIndex = len(hostVolumes) - 1
+
+	hostVolumesEnvVar := corev1.EnvVar{
+		Name:  "HOST_DATA_VOLUME_CONFIG",
+		Value: "",
+	}
+
+	for idx, hostVolume := range hostVolumes {
+		hostVolumesEnvVar.Value += fmt.Sprintf("%s:%s", hostVolume.MountTag, hostVolume.HostPath)
```
tfussell, line 362: Any chance we have a `:` or `,` character in the `MountTag` or `HostPath`?

service/controller/resource/deployment/worker_deployment.go
mazzy89: yes. it would make sense to move somewhere else. I agree here.

service/controller/resource/deployment/worker_deployment.go
mazzy89: good point. if the slice is passed as a reference , all the nested slices will be just value right?

service/controller/key/key.go
mazzy89: an host path is a Unix dir and I do not think it's a common way to name dir containing a comma. meanwhile the `MountTag` is an arbitrary name that we pick here. but I do not know if `,` is accepted value. but for how I've designed the k8s-kvm script, we should not have `,` anywhere otherwise it would fail.

mazzy89:
I need to add here the params that I have added in this PR https://github.com/giantswarm/k8scloudconfig/pull/914

service/controller/resource/deployment/worker_deployment.go
mazzy89: @tfussell where should I move this? which repo?

mazzy89: sure.

mazzy89: done

service/controller/cloudconfig/worker_template.go
```diff
@@ -43,6 +43,14 @@ func (c *CloudConfig) NewWorkerTemplate(ctx context.Context, cr v1alpha1.KVMConf
 		params.SSOPublicKey = c.ssoPublicKey
 		params.ImagePullProgressDeadline = key.DefaultImagePullProgressDeadline
 		params.DockerhubToken = c.dockerhubToken
+		params.KVMWorkerMountTags = []string{}
+
+		if len(cr.Spec.KVM.Workers) > 0 {
+
+			for _, hostVolume := range cr.Spec.KVM.Workers[nodeIndex].HostVolumes {
```
tfussell, line 50:
```suggestion
		if len(cr.Spec.KVM.Workers) > 0 {
			for _, hostVolume := range cr.Spec.KVM.Workers[nodeIndex].HostVolumes {
```

tfussell (approved):

PR #1209, opened 2021-05-20
Proxy flags

yulianedyalkova:
Pass proxy values from flags

### DIFF
--- a/main.go
+++ b/main.go
@@ -129,7 +129,7 @@ func mainError() error {
 
 	daemonCommand.PersistentFlags().String(f.Service.Tenant.Ignition.Path, "/opt/ignition", "Default path for the ignition base directory.")
 	daemonCommand.PersistentFlags().String(f.Service.Tenant.Proxy.HTTP, "", "URL of proxy for HTTP requests.")
-	daemonCommand.PersistentFlags().String(f.Service.Tenant.Proxy.HTTPS, "", "Where to proxy HTTPS trafic to.")
+	daemonCommand.PersistentFlags().String(f.Service.Tenant.Proxy.HTTPS, "", "URL of proxy for HTTPS requests.")
 	daemonCommand.PersistentFlags().StringSlice(f.Service.Tenant.Proxy.NoProxy, []string{}, "List of addresses that need not to go through the proxy.")
 	daemonCommand.PersistentFlags().String(f.Service.Tenant.SSH.SSOPublicKey, "", "Public key for trusted SSO CA.")
 	daemonCommand.PersistentFlags().Bool(f.Service.Tenant.Update.Enabled, false, "Whether updates of tenant cluster nodes are allowed to be processed upon reconciliation.")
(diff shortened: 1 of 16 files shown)


tfussell:
Towards https://github.com/giantswarm/giantswarm/issues/17252

main.go
```diff
@@ -128,6 +128,9 @@ func mainError() error {
 	daemonCommand.PersistentFlags().StringSlice(f.Service.Registry.Mirrors, []string{}, `Image registry mirror domains. Can be set only if registry domain is "docker.io".`)
 
 	daemonCommand.PersistentFlags().String(f.Service.Tenant.Ignition.Path, "/opt/ignition", "Default path for the ignition base directory.")
+	daemonCommand.PersistentFlags().String(f.Service.Tenant.Proxy.HTTP, "", "Where to proxy HTTP trafic to.")
```
tfussell, line 131:
```suggestion
	daemonCommand.PersistentFlags().String(f.Service.Tenant.Proxy.HTTP, "", "URL of proxy for HTTP requests.")
```

main.go
```diff
@@ -128,6 +128,9 @@ func mainError() error {
...
+	daemonCommand.PersistentFlags().String(f.Service.Tenant.Proxy.HTTPS, "", "Where to proxy HTTPS trafic to.")
```
tfussell, line 132:
```suggestion
	daemonCommand.PersistentFlags().String(f.Service.Tenant.Proxy.HTTPS, "", "URL of proxy for HTTPS requests.")
```

service/controller/cloudconfig/cloud_config.go
```diff
@@ -85,10 +87,17 @@ type CloudConfig struct {
 	dockerhubToken  string
 	ignitionPath    string
 	k8sAPIExtraArgs []string
+	proxy           proxyConfig
 	registryMirrors []string
 	ssoPublicKey    string
 }
 
+type proxyConfig struct {
```
tfussell, line 95: Maybe we could drop this and just use `ProxyConfig` (the exported one) in the `CloudConfig` struct?

yulianedyalkova: The exported one has `[]string` type for `NoProxy` where the private one already has the values joined so it's just a string.

tfussell: They could also be joined here https://github.com/giantswarm/kvm-operator/blob/02dc26a6dafc069275ac11a1f8303d9f800675af/service/controller/cloudconfig/master_template.go#L52 or in the template itself, but it's fine if you prefer this way.

tfussell (approved):
LGTM.

service/controller/cloudconfig/cloud_config.go
yulianedyalkova: I will leave it as it is for now cause I don't have time to retest now and we need it asap. I can clean it up in a future commit if that's ok with you :)

PR #1210, opened 2021-05-21
Fix worker host volume indexing

tfussell:


mazzy89 (approved):

tfussell:
We need a follow-up PR so we will test in that one.

PR #1217, opened 2021-05-25
local pv

mazzy89:
Signed-off-by: Salvatore Mazzarino <dev@mazzarino.cz>

### DIFF
--- a/service/controller/resource/pvc/local_pv.go
+++ b/service/controller/resource/pvc/local_pv.go
@@ -46,8 +46,8 @@ func newLocalPVs(customObject v1alpha1.KVMConfig) ([]*corev1.PersistentVolume, e
 					Capacity: map[corev1.ResourceName]resource.Quantity{
 						corev1.ResourceStorage: "",
 					},
-					NodeAffinity: corev1.NodeAffinity{
-						RequiredDuringSchedulingIgnoredDuringExecution: &corev1.NodeSelector{
+					NodeAffinity: &corev1.VolumeNodeAffinity{
+						Required: &corev1.NodeSelector{
 							NodeSelectorTerms: []corev1.NodeSelectorTerm{
 								{
 									MatchExpressions: []corev1.NodeSelectorRequirement{
--- a/service/controller/resource/pvc/local_pv.go
+++ b/service/controller/resource/pvc/local_pv.go
@@ -15,7 +15,7 @@ func newLocalPVs(customObject v1alpha1.KVMConfig) ([]*corev1.PersistentVolume, e
 
 	filesystemVolumeMode := corev1.PersistentVolumeFilesystem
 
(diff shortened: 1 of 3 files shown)


service/controller/resource/pvc/local_pv.go
```diff
@@ -0,0 +1,73 @@
...
+					PersistentVolumeSource: corev1.PersistentVolumeSource{
+						Local: &corev1.LocalVolumeSource{
+							Path: hostVolume.HostPath,
+						},
+					},
+					StorageClassName:              LocalStorageClass,
+					PersistentVolumeReclaimPolicy: corev1.PersistentVolumeReclaimDelete,
+					VolumeMode:                    &filesystemVolumeMode,
+					Capacity: map[corev1.ResourceName]resource.Quantity{
+						corev1.ResourceStorage: "",
```
mazzy89, line 47: The capacity is not enforced here because Kubernetes local storage does not have any way to control the underlying limits (differently by a PVC).

I'm not sure though if it is required. I would vote to remove it but will it fail?

service/controller/resource/pvc/local_pv.go
```diff
@@ -0,0 +1,73 @@
...
+					},
+					NodeAffinity: corev1.NodeAffinity{
+						RequiredDuringSchedulingIgnoredDuringExecution: &corev1.NodeSelector{
+							NodeSelectorTerms: []corev1.NodeSelectorTerm{
+								{
+									MatchExpressions: []corev1.NodeSelectorRequirement{
+										{
+											Key:      "",
```
mazzy89, line 55: Is there a way to pass in here MC masters labels?

service/controller/resource/pvc/resource.go
```diff
@@ -13,6 +13,7 @@ const (
 	// StorageClass is the storage class annotation persistent volume claims are
 	// configured with.
 	StorageClass = "g8s-storage"
+	LocalStorageClass = "local-storage"
```
mazzy89, line 16: We should create beforehand this storage class. I have looked at that but it is not the k8scloudconfig that create `StorageClass`.

service/controller/resource/pvc/local_pv.go
```diff
@@ -0,0 +1,73 @@
...
+					Kind:       "PersistentVolume",
+					APIVersion: "v1",
+				},
+				ObjectMeta: metav1.ObjectMeta{
+					Name: key.LocalPVCName(key.ClusterID(customObject), hostVolume.MountTag),
+					Labels: map[string]string{
+						"app":      key.WorkerID,
+						"cluster":  key.ClusterID(customObject),
+						"customer": key.ClusterCustomer(customObject),
+						"node":     "",
```
mazzy89, line 31: I would put here the MC worker node where the PV points to. Can't say how to pull the MC information in here.

PR #1218, opened 2021-05-26
Add persistant volume claims for local PVs

mazzy89:
Signed-off-by: Salvatore Mazzarino <dev@mazzarino.cz>

### DIFF
--- a/service/controller/key/key.go
+++ b/service/controller/key/key.go
@@ -277,6 +277,10 @@ func EtcdPVCName(clusterID string, vmNumber string) string {
 	return fmt.Sprintf("%s-%s-%s", "pvc-master-etcd", clusterID, vmNumber)
 }
 
+func LocalWorkerPVCName(clusterID string, vmNumber string) string {
+	return fmt.Sprintf("%s-%s-%s", "local-pvc-worker", clusterID, vmNumber)
+}
+
 // FindNodeCondition returns the condition of the given type from the node. The second return value indicates if the condition was found.
 func FindNodeCondition(node corev1.Node, conditionType corev1.NodeConditionType) (corev1.NodeCondition, bool) {
 	for _, condition := range node.Status.Conditions {
@@ -384,15 +388,15 @@ func HostVolumesToVolumeMounts(hostVolumes []v1alpha1.KVMConfigSpecKVMNodeHostVo
 	return volumeMounts
 }
 
-func HostVolumesToVolumes(hostVolumes []v1alpha1.KVMConfigSpecKVMNodeHostVolumes) []corev1.Volume {
+func HostVolumesToVolumes(hostVolumes []v1alpha1.KVMConfigSpecKVMNodeHostVolumes, claimName string) []corev1.Volume {
 	var volumes []corev1.Volume
 
 	for _, hostVolume := range hostVolumes {
 		v := corev1.Volume{
 			Name: hostVolume.MountTag,
 			VolumeSource: corev1.VolumeSource{
-				HostPath: &corev1.HostPathVolumeSource{
-					Path: hostVolume.HostPath,
+				PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{
+					ClaimName: claimName,
 				},
 			},
 		}
@@ -653,6 +657,16 @@ func ShutdownDeferrerPollPath(customObject v1alpha1.KVMConfig) string {
 
 func StorageType(customObject v1alpha1.KVMConfig) string {
 	return customObject.Spec.KVM.K8sKVM.StorageType
+}
+
+func HasHostVolumes(customObject v1alpha1.KVMConfig) bool {
+	for _, worker := range customObject.Spec.KVM.Workers {
+		if len(worker.HostVolumes) > 0 {
+			return true
+		}
+	}
+
+	return false
 }
 
 func ToClusterEndpoint(v interface{}) (string, error) {
--- a/service/controller/resource/deployment/worker_deployment.go
+++ b/service/controller/resource/deployment/worker_deployment.go
@@ -392,7 +392,7 @@ func newWorkerDeployments(customResource v1alpha1.KVMConfig, release *releasev1a
 		}
 		addCoreComponentsAnnotations(deployment, release)
 		// in case of adding additional properties to the worker deployment
-		addConditionalPropsToWorkerDeployment(deployment, capabilities)
+		addConditionalPropsToWorkerDeployment(deployment, customResource, i)
 
 		deployments = append(deployments, deployment)
 	}
@@ -400,7 +400,9 @@ func newWorkerDeployments(customResource v1alpha1.KVMConfig, release *releasev1a
 	return deployments, nil
 }
 
-func addConditionalPropsToWorkerDeployment(deployment *v1.Deployment, caps v1alpha1.KVMConfigSpecKVMNode) {
+func addConditionalPropsToWorkerDeployment(deployment *v1.Deployment, customObject v1alpha1.KVMConfig, workerIndex int) {
+	caps := customObject.Spec.KVM.Workers[workerIndex]
+
 	for i, container := range deployment.Spec.Template.Spec.Containers {
 		if container.Name == key.K8SKVMContainerName {
 			envVars := []corev1.EnvVar{key.HostVolumesToEnvVar(caps.HostVolumes)}
@@ -413,5 +415,6 @@ func addConditionalPropsToWorkerDeployment(deployment *v1.Deployment, caps v1alp
 		}
 	}
 
-	deployment.Spec.Template.Spec.Volumes = append(deployment.Spec.Template.Spec.Volumes, key.HostVolumesToVolumes(caps.HostVolumes)...)
+	claimName := key.LocalWorkerPVCName(key.ClusterID(customObject), key.VMNumber(workerIndex))
+	deployment.Spec.Template.Spec.Volumes = append(deployment.Spec.Template.Spec.Volumes, key.HostVolumesToVolumes(caps.HostVolumes, claimName)...)
 }
--- a/service/controller/resource/pvc/desired.go
+++ b/service/controller/resource/pvc/desired.go
@@ -2,6 +2,7 @@ package pvc
 
 import (
 	"context"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/giantswarm/microerror"
 	corev1 "k8s.io/api/core/v1"
@@ -20,14 +21,38 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
 	if key.StorageType(customObject) == "persistentVolume" {
 		r.logger.Debugf(ctx, "computing the new PVCs")
 
-		PVCs, err = newEtcdPVCs(customObject)
+		etcdPVCs, err := newEtcdPVCs(customObject)
 		if err != nil {
 			return nil, microerror.Mask(err)
 		}
 
+		PVCs = append(PVCs, etcdPVCs...)
+
 		r.logger.Debugf(ctx, "computed the %d new PVCs", len(PVCs))
 	} else {
 		r.logger.Debugf(ctx, "not computing the new PVCs because storage type is not 'persistentVolume'")
+	}
+
+	if key.HasHostVolumes(customObject) {
+		r.logger.Debugf(ctx, "computing the new worker PVCs")
+
+		// Retrieve the existing Persistent Volume in the management cluster to get the storage size
+		// and create the Persistent Volume Claims for the workload cluster's workers
+		pvsList, err := r.k8sClient.CoreV1().PersistentVolumes().List(ctx, metav1.ListOptions{})
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+
+		localPVCs, err := newLocalPVCs(customObject, pvsList)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+
+		PVCs = append(PVCs, localPVCs...)
+
+		r.logger.Debugf(ctx, "computed the %d new PVCs", len(PVCs))
+	} else {
+		r.logger.Debugf(ctx, "not computing the new PVCs because no worker has defined host volumes")
 	}
 
 	return PVCs, nil
--- a/service/controller/resource/pvc/local_pvc.go
+++ b/service/controller/resource/pvc/local_pvc.go
@@ -1,0 +1,57 @@
+package pvc
+
+import (
+	"github.com/giantswarm/apiextensions/v3/pkg/apis/provider/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/giantswarm/kvm-operator/service/controller/key"
+)
+
+func newLocalPVCs(customObject v1alpha1.KVMConfig, pvsList *corev1.PersistentVolumeList) ([]*corev1.PersistentVolumeClaim, error) {
+	var persistentVolumeClaims []*corev1.PersistentVolumeClaim
+
+	for i, workerKVM := range customObject.Spec.KVM.Workers {
+		for _, hostVolume := range workerKVM.HostVolumes {
+			for _, pv := range pvsList.Items {
+				if pv.ObjectMeta.Labels["mount-tag"] == hostVolume.MountTag {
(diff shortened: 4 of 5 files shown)


service/controller/resource/pvc/local_pvc.go
```diff
@@ -0,0 +1,57 @@
...
+							Kind:       "PersistentVolumeClaim",
+							APIVersion: "v1",
+						},
+						ObjectMeta: metav1.ObjectMeta{
+							Name: key.LocalWorkerPVCName(key.ClusterID(customObject), key.VMNumber(i)),
+							Labels: map[string]string{
+								"app":      key.WorkerID,
+								"cluster":  key.ClusterID(customObject),
+								"customer": key.ClusterCustomer(customObject),
+								"node":     "",
```
mazzy89, line 29: Not sure how here passing the Worker ID

service/controller/resource/pvc/desired.go
```diff
@@ -20,15 +21,39 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
...
 	} else {
 		r.logger.Debugf(ctx, "not computing the new PVCs because storage type is not 'persistentVolume'")
 	}
 
+	if key.HasHostVolumes(customObject) {
+		r.logger.Debugf(ctx, "computing the new worker PVCs")
+
+		// Retrieve the existing Persistent Volume in the management cluster to get the storage size
+		// and create the Persistent Volume Claims for the workload cluster's workers
+		pvsList, err := r.k8sClient.CoreV1().PersistentVolumes().List(ctx, metav1.ListOptions{})
```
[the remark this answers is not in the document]
mazzy89: I have added the `LabelSelector` option. I'm not sure if it works.

service/controller/resource/pvc/local_pvc.go
mazzy89: I made an attempt there to pass the worker ID please take a look at it @tfussell

service/controller/resource/pvc/local_pvc.go
```diff
@@ -0,0 +1,63 @@
...
+		workerNode := customObject.Spec.Cluster.Workers[i]
+
+		for _, hostVolume := range workerKVM.HostVolumes {
+			for _, pv := range pvsList.Items {
+				if pv.ObjectMeta.Labels[LabelMountTag] == hostVolume.MountTag {
+					persistentVolumeClaim := &corev1.PersistentVolumeClaim{
+						TypeMeta: metav1.TypeMeta{
+							Kind:       "PersistentVolumeClaim",
+							APIVersion: "v1",
+						},
```
tfussell, line 28: You don't need to define `TypeMeta` as the client manages that during serialization/deserialization.

mazzy89: why then for the `etcd_pvc` is defined in that way?

service/controller/resource/pvc/desired.go
```diff
@@ -20,15 +21,36 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
...
 	} else {
 		r.logger.Debugf(ctx, "not computing the new PVCs because storage type is not 'persistentVolume'")
 	}
 
+	if key.HasHostVolumes(customObject) {
+		r.logger.Debugf(ctx, "computing the new worker PVCs")
+
+		// Retrieve the existing Persistent Volume in the management cluster to get the storage size
+		// and create the Persistent Volume Claims for the workload cluster's workers
+		pvsList, err := r.k8sClient.CoreV1().PersistentVolumes().List(ctx, metav1.ListOptions{LabelSelector: LabelMountTag})
```
tfussell, line 41: I believe the label selector should be of the form `key=value`. `LabelMountTag ` is `mount-tag` which is missing the value part.

mazzy89: I thought that it would work same as `kubectl`. if you run `kubectl get pv -lmount-tag` it works. so then it is required the `key=value`? My assumption is no.

tfussell: Oh right. I just checked the docs and this should work.

service/controller/resource/pvc/local_pvc.go
tfussell: I don't know. It should be fine either way.

service/controller/resource/pvc/local_pvc.go
```diff
@@ -0,0 +1,63 @@
...
+						ObjectMeta: metav1.ObjectMeta{
+							Name: key.LocalWorkerPVCName(key.ClusterID(customObject), key.VMNumber(i)),
+							Labels: map[string]string{
+								"app":      key.WorkerID,
+								"cluster":  key.ClusterID(customObject),
+								"customer": key.ClusterCustomer(customObject),
+								"node":     workerNode.ID,
+							},
+							Annotations: map[string]string{
+								"volume.beta.kubernetes.io/storage-class": LocalStorageClass,
```
tfussell, line 38: I think you should use `storageClassName` instead of this annotation. According to https://kubernetes.io/docs/concepts/storage/persistent-volumes/#class-1, `In the past, the annotation volume.beta.kubernetes.io/storage-class was used instead of storageClassName attribute. This annotation is still working; however, it won't be supported in a future Kubernetes release.`

mazzy89: oh makes a lot of sense. did not check that. then for sure let's change it.

service/controller/resource/pvc/local_pvc.go
```diff
@@ -0,0 +1,63 @@
...
+							Kind:       "PersistentVolumeClaim",
+							APIVersion: "v1",
+						},
+						ObjectMeta: metav1.ObjectMeta{
+							Name: key.LocalWorkerPVCName(key.ClusterID(customObject), key.VMNumber(i)),
+							Labels: map[string]string{
+								"app":      key.WorkerID,
+								"cluster":  key.ClusterID(customObject),
+								"customer": key.ClusterCustomer(customObject),
+								"node":     workerNode.ID,
```
tfussell, line 35: Let's use label keys from `key` package like this https://github.com/giantswarm/kvm-operator/blob/04db84f65e90b9253159bd8b05ab18ad860039bb/service/controller/resource/service/master_service.go#L21-L26

service/controller/resource/pvc/local_pvc.go
```diff
@@ -0,0 +1,63 @@
...
+func newLocalPVCs(customObject v1alpha1.KVMConfig, pvsList *corev1.PersistentVolumeList) []*corev1.PersistentVolumeClaim {
+	var persistentVolumeClaims []*corev1.PersistentVolumeClaim
+
+	for i, workerKVM := range customObject.Spec.KVM.Workers {
+		workerNode := customObject.Spec.Cluster.Workers[i]
+
+		for _, hostVolume := range workerKVM.HostVolumes {
+			for _, pv := range pvsList.Items {
+				if pv.ObjectMeta.Labels[LabelMountTag] == hostVolume.MountTag {
+					persistentVolumeClaim := &corev1.PersistentVolumeClaim{
```
tfussell, line 25: This is 4 levels deep. It might be nicer to find the match and then use it instead of using it inside the `for` loop. You can also negate the `if` condition and use a `continue` so the "happy path" is less indented.

service/controller/resource/pvc/desired.go
```diff
@@ -20,15 +21,36 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
...
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+
+		localPVCs := newLocalPVCs(customObject, pvsList)
+
+		PVCs = append(PVCs, localPVCs...)
+
+		r.logger.Debugf(ctx, "computed the %d new PVCs", len(PVCs))
```
tfussell, line 50:
```suggestion
		r.logger.Debugf(ctx, "computed the %d new worker PVCs", len(PVCs))
```

service/controller/resource/pvc/desired.go
```diff
@@ -20,15 +21,36 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
 	if key.StorageType(customObject) == "persistentVolume" {
 		r.logger.Debugf(ctx, "computing the new PVCs")
```
tfussell, line 22:
```suggestion
		r.logger.Debugf(ctx, "computing the new master PVCs")
```

service/controller/resource/pvc/desired.go
```diff
@@ -20,15 +21,36 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
...
 
-		PVCs, err = newEtcdPVCs(customObject)
+		etcdPVCs, err := newEtcdPVCs(customObject)
 		if err != nil {
 			return nil, microerror.Mask(err)
 		}
 
+		PVCs = append(PVCs, etcdPVCs...)
+
 		r.logger.Debugf(ctx, "computed the %d new PVCs", len(PVCs))
```
tfussell, line 31:
```suggestion
		r.logger.Debugf(ctx, "computed the %d new master PVCs", len(PVCs))
```

service/controller/resource/pvc/local_pvc.go
mazzy89: changed. please check it.

service/controller/key/key.go
```diff
@@ -277,6 +277,10 @@ func EtcdPVCName(clusterID string, vmNumber string) string {
 	return fmt.Sprintf("%s-%s-%s", "pvc-master-etcd", clusterID, vmNumber)
 }
 
+func LocalWorkerPVCName(clusterID string, vmNumber string) string {
```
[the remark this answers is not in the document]
mazzy89: yeah. i noticed that too. just mantained the same pattern to avoid refactoring.

service/controller/resource/deployment/worker_deployment.go
```diff
@@ -392,15 +392,17 @@ func newWorkerDeployments(customResource v1alpha1.KVMConfig, release *releasev1a
 		}
 		addCoreComponentsAnnotations(deployment, release)
 		// in case of adding additional properties to the worker deployment
```
tfussell, line 394: Don't feel like this comment adds any information.
```suggestion
```

service/controller/resource/deployment/worker_deployment.go
```diff
@@ -392,15 +392,17 @@ func newWorkerDeployments(customResource v1alpha1.KVMConfig, release *releasev1a
...
+		addConditionalPropsToWorkerDeployment(deployment, customResource, i)
 
 		deployments = append(deployments, deployment)
 	}
 
 	return deployments, nil
 }
 
-func addConditionalPropsToWorkerDeployment(deployment *v1.Deployment, caps v1alpha1.KVMConfigSpecKVMNode) {
+func addConditionalPropsToWorkerDeployment(deployment *v1.Deployment, customObject v1alpha1.KVMConfig, workerIndex int) {
```
tfussell, line 402: How about renaming this `addHostVolumes` or something since that's all it does?

service/controller/key/key.go
```diff
@@ -384,15 +388,15 @@ func HostVolumesToVolumeMounts(hostVolumes []v1alpha1.KVMConfigSpecKVMNodeHostVo
...
 	var volumes []corev1.Volume
 
 	for _, hostVolume := range hostVolumes {
 		v := corev1.Volume{
 			Name: hostVolume.MountTag,
 			VolumeSource: corev1.VolumeSource{
-				HostPath: &corev1.HostPathVolumeSource{
-					Path: hostVolume.HostPath,
+				PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{
+					ClaimName: claimName,
```
tfussell, line 399: Shouldn't there be a unique PVC name for each PV/host volume? If MC worker 1 has two SSDs, they would both have the same PV name, right? Seems like we need a suffix to tell them apart.

service/controller/resource/pvc/local_pvc.go
```diff
@@ -0,0 +1,64 @@
...
+						},
+						ObjectMeta: metav1.ObjectMeta{
+							Name: key.LocalWorkerPVCName(key.ClusterID(customObject), key.VMNumber(i)),
+							Labels: map[string]string{
+								key.LegacyLabelCluster: key.ClusterID(customObject),
+								key.LabelCustomer:      key.ClusterCustomer(customObject),
+								key.LabelApp:           key.WorkerID,
+								key.LabelCluster:       key.ClusterID(customObject),
+								key.LabelOrganization:  key.ClusterCustomer(customObject),
+								key.LabelVersionBundle: key.OperatorVersion(customObject),
```
tfussell, line 38: We lost the `node` label here. Did you figure out where to get it?

service/controller/resource/pvc/local_pvc.go
mazzy89: I made a small refactor. please take a look.

tfussell: Looking better.

service/controller/resource/pvc/local_pvc.go
tfussell: I could see it being useful. This will make the labels match the deployment labels.

mazzy89: i put it back. please check it.

service/controller/resource/deployment/worker_deployment.go
mazzy89: it adds also the env var. but yeah naming is not the best. whatever thing works for you, works for me.

mazzy89: fixed.

tfussell:
Do we need to check if the PV is already bound to another WC node pod before creating the PVC?

service/controller/key/key.go
mazzy89: oh yeah good catch

mazzy89: fixed.

mazzy89:
with this implementation probably we should because the claim asks for the entire space of the volume so we should check it.

service/controller/resource/pvc/local_pvc.go
```diff
@@ -0,0 +1,79 @@
...
+func findPV(pvsList *corev1.PersistentVolumeList, mountTag string) *corev1.PersistentVolume {
+	for _, pv := range pvsList.Items {
+		if pv.ObjectMeta.Labels[LabelMountTag] != mountTag {
+			continue
+		}
+
+		// discard the PV if is already bound to an existing PV
+		if pv.Spec.ClaimRef != nil {
+			continue
+		}
```
tfussell, line 73: It might be good to move this into `newLocalPVCs` so we can differentiate between "not found" and "already bound" errors.

service/controller/resource/pvc/local_pvc.go
```diff
@@ -0,0 +1,79 @@
...
+			if pv == nil {
+				return nil, microerror.Maskf(notFoundError, "mount tag %s is not available. the mount tag does not exist or is already bound to an existing claim", hostVolume.MountTag)
+			}
+
+			persistentVolumeClaim := &corev1.PersistentVolumeClaim{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: key.LocalWorkerPVCName(key.ClusterID(customObject), key.VMNumber(i), hostVolume.MountTag),
+					Labels: map[string]string{
+						key.LegacyLabelCluster: key.ClusterID(customObject),
+						key.LabelCustomer:      key.ClusterCustomer(customObject),
```
tfussell, line 33: These two are probably not necessary. `LabelOrganization` already gives us the customer and `LegacyLabelCluster` is, well, legacy.

service/controller/resource/pvc/local_pvc.go
```diff
@@ -0,0 +1,79 @@
...
+	LabelMountTag = "mount-tag"
+)
+
+func newLocalPVCs(customObject v1alpha1.KVMConfig, pvsList *corev1.PersistentVolumeList) ([]*corev1.PersistentVolumeClaim, error) {
+	var persistentVolumeClaims []*corev1.PersistentVolumeClaim
+	localStorageClass := LocalStorageClass
+
+	for i, workerKVM := range customObject.Spec.KVM.Workers {
+		for _, hostVolume := range workerKVM.HostVolumes {
+
```
tfussell, line 22:
```suggestion
```

tfussell:
Just a few small comments. I think we're ready to test this.

mazzy89:
I've tested it and it works. ready to release it.

tfussell (approved):

PR #1221, opened 2021-05-28
Fix PVC deletion

tfussell:
The PVs cannot currently be reused after a cluster is deleted because the `ClaimRef` continues to point to the deleted PVC. This PR lists PVCs and removes the `ClaimRef` from bound PVs before deleting them.


service/controller/resource/pvc/current.go
```diff
@@ -19,25 +18,15 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
...
-		if apierrors.IsNotFound(err) {
-			r.logger.Debugf(ctx, "did not find a PVC in the Kubernetes API")
-			// fall through
-		} else if err != nil {
-			return nil, microerror.Mask(err)
-		} else {
-			r.logger.Debugf(ctx, "found a PVC in the Kubernetes API")
-			PVCs = append(PVCs, manifest)
-		}
+	pvcs, err := r.k8sClient.CoreV1().PersistentVolumeClaims(namespace).List(ctx, metav1.ListOptions{
```
tfussell, line 22: Main fix is here. List all cluster PVCs instead of calculating names and getting each by name. This way we pick up local volume PVCs as needing to be deleted so we can unbind them from the PV.

service/controller/resource/pvc/delete.go
```diff
@@ -77,13 +77,13 @@ func (r *Resource) ApplyDeleteChange(ctx context.Context, obj, deleteChange inte
...
 func (r *Resource) NewDeletePatch(ctx context.Context, obj, currentState, desiredState interface{}) (*crud.Patch, error) {
-	delete, err := r.newDeleteChange(ctx, obj, currentState, desiredState)
+	deleteChange, err := r.newDeleteChange(ctx, obj, currentState, desiredState)
 	if err != nil {
 		return nil, microerror.Mask(err)
 	}
 
 	patch := crud.NewPatch()
-	patch.SetDeleteChange(delete)
+	patch.SetDeleteChange(deleteChange)
```
pipo02mix, line 86: I dont get this part, why you do a patch to delete a PVC?

mazzy89 (approved):
The logic looks legit. The fix is on just a few lines. The rest is a lot about renaming and some refactoring. Happy if it has been tested already.

service/controller/resource/pvc/create.go
```diff
@@ -25,8 +25,9 @@ func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange inte
 		r.logger.Debugf(ctx, "creating the PVCs in the Kubernetes API")
 
 		namespace := key.ClusterNamespace(customObject)
-		for _, PVC := range pvcsToCreate {
-			_, err := r.k8sClient.CoreV1().PersistentVolumeClaims(namespace).Create(ctx, PVC, v1.CreateOptions{})
+		for _, persistentVolumeClaim := range pvcsToCreate {
+			toCreate := persistentVolumeClaim
```
yulianedyalkova, line 29: Are you doing this just for the rename? You also might need a deep copy here if you don't want to lose half of the values (arrays for example are not going to be copied afaik).

yulianedyalkova (approved):

service/controller/key/key.go
```diff
@@ -656,7 +646,7 @@ func ShutdownDeferrerPollPath(customObject v1alpha1.KVMConfig) string {
 	return fmt.Sprintf("%s/v1/defer/", ShutdownDeferrerListenAddress(customObject))
 }
 
-func StorageType(customObject v1alpha1.KVMConfig) string {
+func EtcdStorageType(customObject v1alpha1.KVMConfig) string {
```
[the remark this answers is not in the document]
tfussell: If you take a look at https://github.com/giantswarm/kvm-operator/blob/0f3facb6a2a535f40681390ad678030e8442b831/service/controller/resource/pvc/desired.go#L21-L34, you'll see that this storage type is only used by the master deployment and corresponding PVC to determine how to mount the etcd volume. Now that we have different types of storage on masters and workers, I wanted to make it clear that this storage is only the type of the etcd volume and doesn't affect the host volume mount on the workers.

service/controller/resource/pvc/create.go
tfussell: This is mainly to work around a spurious (IMO) warning in `golangci-lint`. Range for loops reuse the same variable on each iteration, so it's usually an error to take the address of the loop variable as in the line below. In this case I don't believe it matters because `r.k8sClient.CoreV1().PersistentVolumeClaims(namespace).Create` doesn't change the object that is passed in. I could have also added `//nolint`. Maybe that would be clearer?

service/controller/resource/pvc/delete.go
tfussell: In our `operatorkit` CRUD resource handlers, a patch includes create, update, and delete steps. So we're not patching the PVC (in the kubernetes sense). Instead, the patch that we're applying is to delete them. Maybe the naming could be improved here, but this particular change was to avoid an editor warning that `delete` is a reserved word in Go.

pipo02mix (approved):

tfussell:
I'll test this in a few different scenarios today.

service/controller/cluster_resource_set.go
```diff
@@ -280,9 +280,9 @@ func newClusterResources(config ClusterConfig) ([]resource.Interface, error) {
 		namespaceResource,
 		serviceAccountResource,
 		configMapResource,
+		pvcResource,
```
tfussell, line 283: Reordering so PVCs are created before deployments.

service/controller/resource/namespace/current.go
```diff
@@ -72,18 +72,29 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
...
-		lo := metav1.ListOptions{
+		deployments, err := r.k8sClient.AppsV1().Deployments(namespace).List(ctx, metav1.ListOptions{
 			LabelSelector: fmt.Sprintf("%s=%s", label.ManagedBy, key.OperatorName),
+		})
+		if err != nil {
+			return nil, microerror.Mask(err)
 		}
 
-		list, err := r.k8sClient.AppsV1().Deployments(n).List(ctx, lo)
+		// Ensure PVCs have been deleted so that bound PVs are properly cleaned up.
```
tfussell, line 84: When a cluster is being deleted, we generally rely on namespace deletion to clean up everything. This was preventing our PVC deletion logic from running.

service/controller/resource/pvc/desired_master.go
```diff
@@ -15,30 +16,29 @@ const (
...
-			},
+		persistentVolumeClaim := corev1.PersistentVolumeClaim{
 			ObjectMeta: metav1.ObjectMeta{
 				Name: key.EtcdPVCName(key.ClusterID(customObject), key.VMNumber(i)),
 				Labels: map[string]string{
-					"app":      key.MasterID,
-					"cluster":  key.ClusterID(customObject),
-					"customer": key.ClusterCustomer(customObject),
-					"node":     masterNode.ID,
+					label.ManagedBy:        key.OperatorName,
```
tfussell, line 32: Taking this opportunity to use labels from `key` instead of hardcoded ones so they match the labels on the worker PVCs.

tfussell:
Re-requesting review since I changed some of the logic. I tested on `gorgoth` and PVs are finally being cleaned up as we expected at the cost of slightly slower cluster deletion.

service/controller/key/key.go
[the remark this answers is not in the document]
yulianedyalkova: I feel like a name of a function should be defined by what it's doing and not by its usage cause that might change in the future and we potentially might have to rename stuff again. But totally fine with leaving it as it is if you think it makes things more readable.

yulianedyalkova (approved):

PR #1236, opened 2021-06-18
Add config to always terminate unhealthy nodes

yulianedyalkova:
Towards https://github.com/giantswarm/giantswarm/issues/17664

Allow unhealthy node termination for all TCs on MC config level. I had 3 ideas of how to do it and this one I maybe hate the least (more ideas welcome). The other two are:
* add new flag to `kubernetesd` and use it to decide if it should put the annotation on the `kvmconfig` or not -> do not like it cause we're leaking the annotation logic to `kubernetesd`
* add a cronjob on our test installations that adds the annotation to all `kvmconfigs` -> we have to run it super often or we can delete the cluster/read the logs before the actual annotating happen and miss the point


helm/kvm-operator/values.yaml
```diff
@@ -44,3 +44,6 @@ registry:
 
 ssh:
   sshPublicKey: ""
+
+unhealthyNodes:
```
yulianedyalkova, line 48: This will be set on config level per installation

service/controller/resource/terminateunhealthynodes/create.go
```diff
@@ -25,8 +25,11 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 
 	// check for annotation enabling the node auto repair feature
 	if _, ok := customResource.Annotations[annotation.NodeTerminateUnhealthy]; !ok {
-		r.logger.Debugf(ctx, "terminate unhealthy node feature is not enabled for this cluster")
-		return nil
+		if !r.alwaysTerminate {
+			r.logger.Debugf(ctx, "terminate unhealthy node feature is not enabled for this cluster")
+			return nil
+		}
+		r.logger.Debugf(ctx, "terminate unhealthy node feature is enabled by default so ignoring annotation")
```
tfussell, line 32:
```suggestion
		r.logger.Debugf(ctx, "terminate unhealthy node annotation not found, enabling feature by default")
```

service/controller/resource/terminateunhealthynodes/create.go
```diff
@@ -25,8 +25,11 @@ func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
 
 	// check for annotation enabling the node auto repair feature
 	if _, ok := customResource.Annotations[annotation.NodeTerminateUnhealthy]; !ok {
-		r.logger.Debugf(ctx, "terminate unhealthy node feature is not enabled for this cluster")
-		return nil
+		if !r.alwaysTerminate {
+			r.logger.Debugf(ctx, "terminate unhealthy node feature is not enabled for this cluster")
```
tfussell, line 29:
```suggestion
			r.logger.Debugf(ctx, "terminate unhealthy node annotation not found, skipping reconciliation")
```

service/controller/resource/terminateunhealthynodes/create.go
yulianedyalkova: Updated.

service/controller/resource/terminateunhealthynodes/create.go
yulianedyalkova: Updated.

helm/kvm-operator/templates/configmap.yaml
```diff
@@ -56,3 +56,5 @@ data:
           ssoPublicKey: '{{ .Values.ssh.ssoPublicKey }}'
         update:
           enabled: true
+      unhealthyNodes:
+        alwaysTerminate: '{{ .Values.unhealthyNodes.alwaysTerminate }}'
```
pipo02mix, line 60: What about name it like AWS `terminate-unhealthy` or `terminate-unhealthy-nodes` and the use here the param `enabled` equal to `true|false` as we do with others?

pipo02mix: It means the flag name will change it too 🙈

main.go
```diff
@@ -133,6 +133,7 @@ func mainError() error {
 	daemonCommand.PersistentFlags().StringSlice(f.Service.Tenant.Proxy.NoProxy, []string{}, "List of addresses that need not to go through the proxy.")
 	daemonCommand.PersistentFlags().String(f.Service.Tenant.SSH.SSOPublicKey, "", "Public key for trusted SSO CA.")
 	daemonCommand.PersistentFlags().Bool(f.Service.Tenant.Update.Enabled, false, "Whether updates of tenant cluster nodes are allowed to be processed upon reconciliation.")
+	daemonCommand.PersistentFlags().Bool(f.Service.UnhealthyNodes.AlwaysTerminate, false, "Whether to terminate unhealthy nodes on all WCs by default.")
```
pipo02mix, line 136:
```suggestion
	daemonCommand.PersistentFlags().Bool(f.Service.UnhealthyNodes.AlwaysTerminate, false, "Whether or not to terminate unhealthy nodes on all WCs by default.")
```

pipo02mix (commented):
minor comments

helm/kvm-operator/templates/configmap.yaml
tfussell: `terminateUnhealthyNodes` feels a little clearer, but I don't feel strongly if you want to keep it.

tfussell (approved):

helm/kvm-operator/templates/configmap.yaml
yulianedyalkova: I find it a bit misleading if it's just enabled/disabled. Because when I see it disabled I would assume that it disables the whole feature and will not terminate nodes even when the annotation is set and that's not the case. It's the other way around - if you have this enabled it will ignore the annotation and always terminate. If it's disabled it will default to using the annotation only.

yulianedyalkova: Renamed it to `terminateUnhealthyNodes`

PR #1250, opened 2021-07-19
Change msize to 512000

yulianedyalkova:
Towards https://github.com/giantswarm/giantswarm/issues/17377

`512000` (~500kib) is the max value that the kernel currently supports. More context [here](https://github.com/giantswarm/giantswarm/issues/17377#issuecomment-882268970).


pipo02mix (approved):
It sounds rasonable, good improvement!

PR #1312, opened 2022-01-24
Check all deployment pod statuses before updating deployments

tfussell:


tfussell:
Closing in favor of https://github.com/giantswarm/kvm-operator/pull/1313

PR #1337, opened 2022-07-04
Parse registry mirrors properly

pipo02mix:


helm/kvm-operator/templates/configmap.yaml
```diff
@@ -48,7 +48,10 @@ data:
         {{- end }}
       registry:
         domain: '{{ .Values.registry.domain }}'
-        mirrors: '{{ range $i, $e := .Values.registry.mirrors }}{{ if $i }},{{end}}{{ $e }}{{end}}'
+        mirrors: 
+        {{- range $i, $e := .Values.registry.mirrors }}
+        - {{ $e | quote }}
+        {{- end }}
```
kopiczko, line 54: Maybe it makes sense to protect agains empty mirrors like here https://github.com/giantswarm/config/blob/5f123cdc4d419f6be246a952966ca331955e43ff/default/apps/kvm-operator/configmap-values.yaml.template#L19-L25

kopiczko (approved):
LGTM

helm/kvm-operator/templates/configmap.yaml
pipo02mix, line 54:
```suggestion
        {{ if .registry.mirrors }}
        mirrors: 
        {{- range $i, $e := .Values.registry.mirrors }}
        - {{ $e | quote }}
        {{ else }}
        mirrors: []
        {{ end }}
```

PR #1344, opened 2022-07-06
Fix extra space chart2

pipo02mix:


pipo02mix:
This is sad, but helm template command does not get these errors :( I am testing the chart directly right now

helm/kvm-operator/templates/configmap.yaml
```diff
@@ -49,13 +49,13 @@ data:
...
         {{- range $i, $e := .Values.registry.mirrors }}
-        - {{- $e | quote -}}
-        {{ end }}
-        {{ else }}
+        - {{ $e | quote -}}
+        {{ end -}}
+        {{ else -}}
         mirrors: []
-        {{ end }}
+        {{- end }}
```
kopiczko, line 58: I can't guarantee that but this should be correct:

```
        {{- if .Values.registry.mirrors -}}
        mirrors:
        {{- range $i, $e := .Values.registry.mirrors }}
        - {{ $e | quote -}}
        {{- end }}
        {{- else }}
        mirrors: []
        {{- end }}
```

But feel free to ignore if you tested your version already

kopiczko (approved):

helm/kvm-operator/templates/configmap.yaml
pipo02mix: This configuration works, just installed from control plane test catalog. Thanks!

PR #1359, opened 2023-05-10
Consider new control-plane taint

mogottsch:
This PR considers the new taint `node-role.kubernetes.io/control-plane` when determining whether a node is unschedulable.

Towards https://github.com/giantswarm/roadmap/issues/2468


mogottsch:
Closed because the KVM platform is EOL and won't ever be upgraded to the problematic Kubernetes version.

PR #1363, opened 2023-06-12
disable renovate

glitchcrab:
towards https://github.com/giantswarm/giantswarm/issues/26446


vxav (approved):
My thumb is pointing upwards

